Issue #18, opened 2023-12-14
envvar on Parameter()


rbonthond:
would be nice if we could support getting cli values via environment variables (https://typer.tiangolo.com/tutorial/arguments/envvar/)

BrianPugh:
I can add this feature soon 👍

BrianPugh:
This is now available in v1.1.0

PR #20, opened 2023-12-14
Feature: Add environment variable parsing.

BrianPugh:
Allows for environment variable parsing fallbacks:

```
@app.default
def foo(bar: Annotated[int, Parameter(env_var="BAR")] = 123):
    pass
```

The priority order is:
1. CLI-provided values
2. Values parsed from `Parameter.env_var`.
3. Default values from the function signature.

Addresses #18.

### DIFF
--- a/cyclopts/bind.py
+++ b/cyclopts/bind.py
@@ -1,4 +1,5 @@
 import inspect
+import os
 import shlex
 import sys
 from functools import lru_cache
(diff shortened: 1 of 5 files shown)


BrianPugh:
@rbonthond let me know what you think of this.

rbonthond:
that works for me, thank you!

Issue #24, opened 2023-12-18
Ability to set global overrides for `cyclopts.Parameters`


Ravencentric:
Take this case as an example:
I do not want negatives or defaults to show up in --help for everything but one. Currently, I gotta do something like:
```py
foo: Annotated[bool, Parameter(name="--foo", help="do foo", negative="", show_default=False)] = False,
bar: Annotated[bool, Parameter(name="--bar", help="do bar", negative="", show_default=False)] = False,
baz: Annotated[bool, Parameter(name="--baz", help="do baz")] = False,
```
Is it (or would it be) possible to set some global default overrides? 
Example:

```py
cyclopts.GlobalParameters(negative="", show_default=False)

foo: Annotated[bool, Parameter(name="--foo", help="do foo")] = False,
bar: Annotated[bool, Parameter(name="--bar", help="do bar")] = False,
# override show_default global for baz but negative continues to follow global default
baz: Annotated[bool, Parameter(name="--baz", help="do baz", show_default=True)] = False,
```

BrianPugh:
I'm actually already working on this feature! I think it's going to look like this:

```
from cyclopts import App, Parameter

app = App(default_parameter=Parameter(negative="", show_default=False))

# Now all Parameters have `negative=""` and `show_default=False` as the default.
```

Any feedback on this proposed solution?

Ravencentric:
Looks perfect, thank you!

BrianPugh:
This might have to wait a week or two. There's a [newly merged feature in attrs](https://github.com/python-attrs/attrs/pull/1187) that significantly simplifies the implementation. [Hynek intends to do a release before new years](https://github.com/python-attrs/attrs/issues/1212#issuecomment-1858940052).

BrianPugh:
I found a workaround via the [`record_init_kwargs` decorator](https://github.com/BrianPugh/cyclopts/pull/33/files#diff-843c8548fd7abe5479a26f75ba52b1c4443c3014d18821dbdcd6e7f830bc5fbeR4).

BrianPugh:
This has been and released in v1.2.0.

PR #33, opened 2023-12-20
Default parameter overrides.

BrianPugh:
This ended up being a super chunky PR, but here's the summary of changes:

* `App` now takes an optional parameter `default_parameter: Parameter`.  Parameters now have a resolution order for determining configuration values. Basically it goes (highest-to-lowest priority)`annotated parameter` -> `parenting app default` -> `parenting-parenting app default` -> `...`.
    * **BREAKING** (but not in a meaningful way): removed `MultipleParameterAnnotationError`; we now handle multiple Parameter resolution. Specifically, for mutliple `Parameter` in an `Annotated`, they are evaluated left->right (right-most has highest precedence/priority).
* `App` and `Parameter`'s repr strings have been greatly simplified to only include non-default supplied parameters.

* **BREAKING** (but not in a meaningful way): all of `Parameter`'s defaults are now `None`. No change in default functionality.
    * New classmethod `Parameter.combine` which can construct a new, single, Parameter from multiple Parameters.
    * New classmethod `Parameter.default` which is similar to `Parameter()`, but it will override all parenting Parameters.

This will be a part of v1.2.0

Addresses #24.

@Ravencentric I intend to merge this over the next day, but I'd appreciate it if you could play around with it (now onto addressing your other Github issues 🙈 )


Ravencentric:
I'll check this out tonight!

Ravencentric:
Okay, I got around to testing it and didn't manage to find any issues!

PR #64, opened 2024-01-13
Group sort key

BrianPugh:
Adds a new attribute ``Group.sort_key`` that influences the order groups are displayed on the help-page.


Ravencentric:
Tried this! Didn't find any issues!

Issue #65, opened 2024-01-13
Path validator gives misleading error message


Ravencentric:
take this for an example
```py
# Must exist, must be a file, must NOT be a directory
validator=validators.Path(exists=True, file_okay=True, dir_okay=False)
```
```txt
> python -m test --config "D:\this_dir_exists"
╭─ Error ──────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Invalid value for --config. D:\this_dir_exists is not a directory.                                                   │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```
This is wrong, It is a directory, it's not supposed to be a directory but it is one. same with `file_okay=False`

The message should be something like `Invalid value for --config. Only file input is allowed but D:\this_dir_exists is a directory.`

BrianPugh:
Thanks for the report! To confirm, the solution is just to remove the "not" from the error message, correct? This one melted my brain because it has like 3 negations 😄 .

If so, this will be fixed in develop-v2 (eta still by EOD sunday).

Ravencentric:
> the solution is just to remove the "not"
Yep! Currently it says the exact opposite of what it should

BrianPugh:
will close this issue once v2 is released. I'll let this sit for another day in case you discover any behavior/bugs you don't like.

I'll also try and add some more tests for any lines that currently don't have coverage.

Ravencentric:
I have been thinking about this and I personally don't find `exist=False` a good default. I feel like if I'm going to use the validator without any explicit arguments like,`validator=validators.Path()`, then I'm expecting it to atleast check it's an existing path on the system.

BrianPugh:
hm, good to have this discussion now in-case we want to make a breaking change.

1. I agree, I think `validators.Path(exist=True)` is the typical use, and so for that reason would make a better default.

2. The current default behavior is "least restrictive", and can be configured to be more restrictive.

3. The `click` (and, by extension, Typer) [default is `exist=False`](https://click.palletsprojects.com/en/7.x/parameters/#parameter-types). Not that this is justification alone, but should be additionally considered.

For reason (1), I want to change it. For reason (2), I want to keep it as-is.

BrianPugh:
An additional option, not-mutually-exclusive-option: I introduce a new submodule `cyclopts.types` that just has a bunch of annotated common types.

So for this use-case:

```
[...]

PR #67, opened 2024-01-13
Fix validators.Path error messages.

BrianPugh:
Addresses #65


BrianPugh:
@ravencentric what do you think of this?

Ravencentric:
Perfect, the extra post init check is also a very welcome change

Issue #69, opened 2024-01-14
Ability to modify default Usage string


Ravencentric:
There should be `usage` argument in `App()` that I could use to modify the default usage string in `--help`

```
> python -m juicenet --help
Usage: juicenet COMMAND [ARGS] [OPTIONS] <---- Change this

CLI tool designed to simplify the process of uploading files to Usenet

╭─ Arguments ──────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ PATH  file or directory. [default: C:\Users\raven\Documents\GitHub\juicenet-cli]                                     │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Parameters ─────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --config            path to your juicenet config file [env var: JUICENET_CONFIG]                                     │
│ --public            use your public/secondary nyuu config                                                            │
│ --nyuu              only run nyuu                                                                                    │
│ --parpar            only run parPar                                                                                  │
│ --raw               only repost raw articles                                                                         │
│ --skip-raw          skip raw article reposting                                                                       │
│ --clear-raw         delete existing raw articles                                                                     │
│ --exts              file extensions to be matched, overrides config                                                  │
│ --glob              glob pattern(s) to be matched instead of extensions                                              │
│ --bdmv              search for BDMVs in path, can be used with --glob                                                │
│ --debug             show debug logs                                                                                  │
│ --move              move foobar.ext to foobar/foobar.ext                                                             │
│ --only-move         move foobar.ext to foobar/foobar.ext and exit                                                    │
│ --no-resume         ignore existing resume data                                                                      │
│ --clear-resume      delete existing resume data                                                                      │
│ --help          -h  display this message and exit                                                                    │
│ --version           display application version                                                                      │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

BrianPugh:
Ah yes, I forgot about usage.

I'll do this two-fold:

1. Improve the current usage to be more consistent with other libraries (the current was a bit of a lazy place holder).

2. Allow manual override via a supplied string `usage="My Usage"`.

3. Allow disabling it via supplying an empty string.

Skyler-Altol:
@BrianPugh would it not be more consistent with the general patterns of python (and this project) to use None for the disabling component? Just a thought :) 

[...]

PR #71, opened 2024-01-15
Add new App.usage field.

BrianPugh:
Partially addresses #69.  Allows a manually specified `App.usage` string.

I'm postponing improving the auto-generated usage string for a minor v2.X release in a future PR; this is a relatively small-value feature improvement and I don't want to block v2 release on it.

@ravencentric please give this a look!


Ravencentric:
I don't want v2 to be delayed either and this looks good as well!

Issue #77, opened 2024-01-17
support for union of None like Optional


samsja:
Hey, thanks for the great lib.

I think that currently  `Optional[int]` is supported but not `int | None` which should be equivalent.

Would it be possible to support it ?

thanks in advance

BrianPugh:
Thanks for the bug report! Fixing now. I mostly code in python3.8 syntax (until it reaches EOL), so I didn't notice this issue.

The problem at heart is basically described in https://github.com/python/cpython/issues/105499

BrianPugh:
Should be fixed in v2.1.1

samsja:
Thanks for the ultra quick fix ! Just updated to the latest version and it is working.

> The problem at heart is basically described in https://github.com/python/cpython/issues/105499

okay this is really weird  problem that Union and | don't have the same behavior, very counterintuitive but good to know :thinking:

PR #80, opened 2024-01-17
Support python310 pipe unions.

BrianPugh:
Addresses #77


BrianPugh:
@samsja can you give this a try and see if it resolves the bug you found?

PR #112, opened 2024-02-23
Added async functions support

nesb1:
Hello!

Cyclopts is a beatiful cli library!

I just started use it and found it necessary to have async support, because a lot of cli-apps doing async work. 

Now we have an option to extract async code to function and call it with `asyncio.run`, it is not comfortable.

So, I propose to do it instead of user.


BrianPugh:
thank you very much for this! I'll be making some minor tweaks to it this weekend, then merge & cut a release (eta by sunday).

BrianPugh:
This is included in the newly released v2.4.0

PR #165, opened 2024-04-30
Config command defaults (e.g. from toml or yaml file)

### LINKED ISSUE #87: [Discussion] First-class config overrides via pyproject.toml?
Hi! Thanks for building this awesome tool. I'm loving using it so far, and excited to show our users how awesome our CLIs can be. I did want to get your thoughts on how you might envision cyclopts evolving. There are some areas the story seems a little unclear with how configs loaded from pyproject.toml interact with those passed in. These aren't feature requests, but it would be good to get your reaction to these, and the general idea of first-class support for pyproject.toml:

### Meta App Parameter Overrides

While there is a nice example showing how to override values for commands, I was playing around with this a bit and didn't see an immediately obvious way to infer defaults for parameters passed to the meta app. Take the following example, using verbose and quiet logging flags as an example:

```python
log_params = Group(
    "Logging",
    default_parameter=Parameter(
        negative="",  # Disable "--no-" flags
        show_default=False
    ),
    validator=validators.LimitedChoice(),  # Mutually Exclusive Options
)

# This method requires these types to be optional even if they're not
VERBOSE = Annotated[bool | None, Parameter(help="Increase log output", group=log_params)]
QUIET = Annotated[bool | None, Parameter(help="Decrease log output", group=log_params)]

# Separate dict to track actual defaults since we have to have the CLI default to None
DEFAULTS = {
    'verbose': False,
    'quiet': False
}

def get_meta_param(name: str, value: Any, config: dict[str, Any]) -> Any:
    """Get parameter for meta app from config, accounting for pyproject.toml and command line overrides.

    All meta app parameters must be optional/nullable in order to sue this as a flag for whether or not
    they have been set via the command line.

    Args:
        name: Name of the parameter
        value: Value of parameter received from CLI
        config: Config dict from pyproject.toml

    Returns:
        Parameter value
    """
    # Flag was passed via CLI
    if value is not None:
        return value

    # Flag was specified in pyproject.toml
    if config.get(name, None) is not None:
        return config[name]
    
    # Flag was specified in neither, set true default
    return DEFAULTS[name]


@app.meta.default
def meta(
    verbose: VERBOSE = None,
    quiet: QUIET = None
) -> Any:
    """Meta App. Used to inject configs from pyproject.toml as defaults
    """
    config = load_pyproject_config()

    # The main Cyclopts parsing/conversion
    command, bound = app.parse_args(tokens)

    # Example of using these - the internals here don't matter
    configure_logging(
        verbose=get_meta_param('verbose', verbose, config),
        quiet=get_meta_param('quiet', quiet, config)
    )

    # Other config parsing for commands below here
    ...
```

[edit: in case it's not clear why I felt like I have to use `None` as a default, we otherwise couldn't tell the difference between a passed False config and the method default value of False]

It is entirely possible I am way off the prescribed path here, so if there is an easier way to do this, please let me know. If not, it should be hopefully a little clear how using meta app params ends up feeling a little complex and also locks you out of documenting default values via type hints.

### Config Validation

Configs loaded via pyproject.toml seem to skip the standard validation for CLI parameters. E.g. I could have updated the defaults via pyprject to some invalid combination, and this would not be caught by e.g. `validators.LimitedChoice()`. You can test out the example above, by adding a section to your project.yaml with both `verbose` and `quiet` config set to true.

There's probably a number of ways to work around this, but I think you'd have to load in the pyproject.toml early to get these values set in advance of the current validator logic. Perhaps loading pyproject.toml on init and attaching default_config dict to the App class would be one way to go about this? There are probably some considerations for not accidentally changing the defaults as they show in the CLI help text.

### Schema Validation

It would be really awesome to be able to generate schema validation json for any commands. By this, I mean some compatibility functionality to generate a schema.json file from the command definitions themselves. These schema definitions can be used by IDEs etc to provide inspection capability to users who are adding a section to configure a published tool. An example of ruff's schema.json is https://github.com/astral-sh/ruff/blob/main/ruff.schema.json.

Having a defined schema for pyproject.toml input would also allow auto-coercion for any types from pyproject.toml, so you wouldn't have to manually validate each config.

Thanks again for sharing this library, it really is great, and I'd love to hear more about your thoughts on this problem space.

BrianPugh:
This PR aims to make it easier to supply configuration defaults from a configuration file by introducing a new `cyclopts.App` parameter called `bound_args_transform`.

`bound_args_transform` is a callable (or list of callables) that has signature:
```python
 def bound_args_transform(commands: Tuple[str, ...], bound: inspect.BoundArguments) -> Any:
     ...
```

The supplied functions should add/modify the provided `bound` object based on some external configuration. The `commands` parameter are the string CLI commands/subcommands that lead up to the function-to-be-executed. For example,
```bash
# assuming "cmd" and "subcmd" are cyclopts commands.
$python my-script.py cmd subcmd --foo=bar
```
would result in `commands == ("cmd", "subcmd")`

Cyclopts has a few builtin transform classes for common situations for yaml, toml, and json. Example usage:
```python
from cyclopts import App, bound_args_transforms

app = App(
    bound_args_transform=bound_args_transforms.Toml("pyproject.toml", root_keys=["tool", "mytoolname"])
)
```
Extending the above example, this would:
1. Search for pyproject.toml in the current directory. If it's not found, do not modify the `bound` object at all.
2. If it does exist, look inside for the table `[tool.mytoolname.cmd.subcmd]` and supply the key/value pairs for unset parameters.

Other nice parameters of the builtin transforms:
* `must_exist=False`: whether or not the configuration file must exist.
* `search_parents=False`: iteratively search parenting directories until a config file is found.

Resolves #87. @BrendanJM what do you think about the concepts outlined in this post? This branch is not in a functional state yet (and the implementation is not yet ready to be reviewed), but I was wondering if I could get your high level feedback?


BrendanJM:
Hi Brian - this looks great! Looking through this, a few thoughts:
- I like that various config formats could be used. Though focusing on only e.g. `pyproject.toml` seems like it could also be valid and idiomatic.
- When searching for e.g. `pyproject.toml` in the current directory, is that relative to where the command is run? I imagine looking for this in the repo root by default would potentially be desirable (and maybe obviate the need to search parents). Is there conventional process for locating this (e.g. what setuptools uses, etc)?
- [nit] The interface looks good, but to a new/basic user, `bound_args_transform` may benefit from a more descriptive/simplified name. But this also isn't a big deal given decent documentation and examples.

BrianPugh:
Thanks for looking!

> Though focusing on only e.g. pyproject.toml seems like it could also be valid and idiomatic.

I kinda agree, however there are many ways to interpret `pyproject.toml` that may be project-specific. The default implementation included in Cyclopts is probably generically useful, but people may want to tweak it for their project. The natural conclusion is this function-callable interface where the user can parse things however they would like.

> When searching for e.g. pyproject.toml in the current directory, is that relative to where the command is run?

Yes. But again, this is project specific. For example, if we write a CLI tool that lints a repo based  on it's `pyproject.toml`, we would want:

```
app = App(
    bound_args_transform=bound_args_transforms.Toml("pyproject.toml", root_keys=["tool", "mytoolname"], search_parents=True)
)
```

However, some other tools may want to look in some home config directory:
```
app = App(
    bound_args_transform=bound_args_transforms.Toml("~/.config/mytoolname/config.toml")
)
```

In either case, it's pretty explicit about where the configuration will be coming from.

> `bound_args_transform` may benefit from a more descriptive/simplified name.

I 100% agree, but I couldn't think of a better name. Totally open to suggestions! A odd nuance/coincidence is that this hook is VERY similar to `App.converter`, just with a slightly different interface. It might be worth combining them in a non-backwards compatible way.

BrianPugh:
After a little more time, i have the following thoughts:

* This should NOT be merged with `App.converter`, since `App.converter` is NOT inherited from the parenting app. This new feature should be inherited all the way through.
* I think we should give this the much simpler parameter name `config`, since that's the primary use-case. It'll also be much easier for people to find the feature.

BrendanJM:
Agree - I think the simple `config` is a great name here, and would be a lot more clear in terms of intent.

PR #172, opened 2024-05-15
Improve type inference when using decorators with pyright

breathe:
## Problem

`pyright` infers the wrong types for functions which are decorated with `@app.command` or `@app.default`.

Use of `TypeVar` here allows to explicitly preserve the original type information of the original non-wrapped function argument which isn't modified in anyway by the decorators


BrianPugh:
released with v2.6.2, thank you for the contribution!

PR #194, opened 2024-07-04
Changed the type-hints of App.command() to support recent versions of pyright

andersjel:
This makes two changes:

* Adds a type hint to kwargs. This will prevent recent versions of pyright from reporting an error every time App.command() is used about partially unknown argument types.
* Adds type hints for the two ways App.command() can be used using typing.overload().

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 exclude: ^(poetry.lock|.idea/)
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: "v0.4.8"
+    rev: "v0.5.0"
     hooks:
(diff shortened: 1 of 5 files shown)


BrianPugh (approved):

BrianPugh:
Sorry to muck up your PR, but I had to add a bunch of fairly unrelated things to get CI to pass. Let me know if you have any thoughts/alternative solutions to my commit. A key annoying issue is that pyright doesn't know about `attrs` converters for attributes, so we have to type hint them as everything that they accept, rather than what they are after conversion.

andersjel:
> Sorry to muck up your PR, ...

No worries, I am happy you took up the PR.

I do not have any alternative solutions.

PR #220, opened 2024-08-21
Use last command's help/version flags.

### LINKED ISSUE #219: the --help and --version parameters seems to not get parsed correctly because they are default commands
this is simplified version of the code i wrote

wrote some commands that build some kind of debian package
one parameter was optional version if it wasnt passed i would detect it from github/config file 
when i passed version i wanted to insert the passed version to the debian package instead to skip the detecting step
it seemed that --version flag wasnt working so i wrote this small command to see the generated control file
for debugging and simplified it for this example

```python3
import logging

from cyclopts import App,Parameter
app=App()
from typing import Annotated
@app.command(name="print-control-file")
def print_control(version:Annotated[str|None,Parameter(help="control file version")]):
    x=f"""Package: <package name>
Version:{version}-release
Architecture: amd64
Section: admin
Priority: optional
Depends: python3,lsof
Recommends:
"""
    print(x)
def main():
    logging.basicConfig(level=logging.DEBUG, format="%(name)s - %(levelname)s - %(message)s", stream=sys.stderr)
    app()


````

```bash
 > ./cli.py  print-control-file --version 1.2.3 -h
Usage: cli.py print-control-file [ARGS] [OPTIONS]

╭─ Parameters ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ *  VERSION,--version  control file version [required]                                                                             │
╰─────────────────────────────────────────────────────────────
```

```bash
./cli.py  print-control-file --version 1.2.3
0.0.0
````

```bash
./cli.py  print-control-file --version=1.2.3
Package: <package name>
Version:1.2.3-release
Architecture: amd64
Section: admin
Priority: optional
Depends: python3,lsof
Recommends:
```

i tried to check overriding --help
and i got same behaviour

so  i thought thats bug either in allowing such flags to be registered or by 
parsing the command args

checked 0 parameters commands but those worked so i think it only for default commands that start with --help --version or -h but i might be wrong

for example i tried( and it worked)
```python3
from cyclopts import App,Parameter
app=App()
from typing import Annotated

@app.command(name="--test")
def test():
    print('--test was called')

@app.command(name="print")
def print_test(test:Annotated[str,Parameter(name="--test")]="value"):
    if test:
        print("print_test was called",test)
def main():
    logging.basicConfig(level=logging.DEBUG, format="%(name)s - %(levelname)s - %(message)s", stream=sys.stderr)
    app()


if __name__ == "__main__":
    main()
```
after that i checked with typer to see if/how it behave in such cases.
( it dont have --version  param or -h by default)
so i checked --help param overriding
and typer handled that one correctly
```python3

import typer
from typer import Option
app=typer.Typer(name="test")

@app.command(name="print2")
def print_help_test_2(help_msg:Annotated[str,Option("--help")]="help-message-content"):
    print("test command was called with msg:",help_msg)
    

@app.command(name="print")
def print_help_test(print_help:Annotated[bool,Option("--help")]=False):
    if print_help:
        print("--help was passed")
    else:
        print("--help wasn't passed")

def main():
    app()
if __name__ == "__main__":
    main()

````

```bash
 > ./cli.py print --help
--help was passed

 > ./cli.py print2 --help help-value
```

BrianPugh:
Kind of resolves #219.

The issue was that if a command wants to use a parameter like `--version` or `--help`, the version/help flag system swallows it and gets priority. Previously, only the help/version flags of the root app were used; now the help/version flags of the resolved command are used (more intuitive/correct).

@orhayat with this PR, you can use `version` in your command as the following:

```python
# notice the `version_flags`
@app.command(name="print-control-file", version_flags=[])
def print_control(version:Annotated[str|None,Parameter(help="control file version")]):
    x=f"""Package: <package name>
Version:{version}-release
Architecture: amd64
Section: admin
Priority: optional
Depends: python3,lsof
Recommends:
"""
    print(x)
```

I think I would like for the command to automatically have precedence over the help/version flags, but I don't think that will be possible until cyclopts v3.

Keeping this as a draft as I'm not totally satisfied with the solution; I want to think about it a little more.

### DIFF
--- a/cyclopts/core.py
+++ b/cyclopts/core.py
@@ -5,6 +5,7 @@ import traceback
 from contextlib import suppress
 from copy import copy
 from functools import partial
+from itertools import chain
 from pathlib import Path
 from typing import (
     TYPE_CHECKING,
@@ -291,8 +292,10 @@ class App:
 
     def __attrs_post_init__(self):
         # Trigger the setters
-        self.help_flags = self._help_flags
-        self.version_flags = self._version_flags
+        func = getattr(self.default_command, "__func__", None)
+        if func != type(self).version_print and func != type(self).help_print:
+            self.help_flags = self._help_flags
+            self.version_flags = self._version_flags
 
     ###########
     # Methods #
@@ -311,7 +314,7 @@ class App:
         for command in commands:
             with suppress(KeyError):
                 if default:
-                    if self[command].default == self.version_print:
+                    if self[command].default == default:
                         del self[command]
                 else:
                     del self[command]
@@ -324,14 +327,18 @@ class App:
     def version_flags(self, value):
         self._version_flags = value
         self._delete_commands(self._version_flags, default=self.version_print)
-        if self._version_flags:
+        func = getattr(self.default_command, "__func__", None)
+        if self._version_flags and func != type(self).version_print:
+            assert isinstance(self._version_flags, tuple)
             self.command(
-                self.version_print,
-                name=self._version_flags,
-                help_flags=[],
-                version_flags=[],
-                version=self.version,
-                help="Display application version.",
+                App(
+                    name=self._version_flags,
+                    default_command=self.version_print,
+                    help_flags=self.help_flags,
+                    version_flags=self.version_flags,
+                    version=self.version,
+                    help="Display application version.",
+                )
             )
 
     @property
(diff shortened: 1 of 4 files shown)


BrianPugh:
alright, so I have decided to go forward with this. Worst case, it's more intuitive before and allows for the very reasonable use-case in #219.

In cyclopts v3 I will probably try to have the command's parameters get priority over version/help-flags.

BrianPugh:
there is a bit of ??? around which app's `version_print` and `help_print` should be used, but I think we can punt on that issue for now.

PR #244, opened 2024-11-08
fix: readme grammar

YodaDaCoda:
it's == "it is"

apologies for being pedantic


BrianPugh:
oh man, I knew my grammar isn't the best, but TIL. I think i'll have to make this change elsewhere in the docs. Thanks!

Issue #248, opened 2024-11-12
Feature request: Automatic grouping for nested objects


vhdirk:
Hi! First of, thanks for this amazing library. I've been eagerly awaiting v3!

I have an application with a fairly large set of options, that I've set up like this:

```python


@dataclass
class SerialOptions:
    device: CharDevice = Path("/dev/ttyUSB0")
    baudrate: Annotated[int, Parameter(name=["--rate"])] = 9600


@dataclass
class Tester:
    """Perform tests.

    Args:
        test_suite: The test suite to run
        serial: serial device options
        test_variant: define variant for tests
        test_mods: modifications for all test suites and variants
        device: options for device
        timeout: Maximum timeout, defined in minutes
    """

    test_suite: Annotated[TestSuite, Parameter(name=["--test"])] = TestSuite.GENERAL
    serial: SerialOptions | None = None
    test_variant: TestVariant | None = None
    test_mods: Annotated[set[TestMod] | None, Parameter(name=["--mod"])] = None
    device: Annotated[
        DeviceOptions | None, Parameter(name=["--external-flash", "--flash"])
    ] = None
    timeout: int = 40

    ...



app = App(name="tester")

@app.default
def main(tester: Annotated[Tester, Parameter(name="*")]):
    tester.run()


if __name__ == "__main__":
    app()

```

What would be nice is that stuff like `serial` would automatically be considered a `Group`. Since that is exactly why I structured it this way :)

vhdirk:
Thanks for the fast response!

I like the idea of the extra parameter. Alternative, or on top of that, what about a `group` decorator on the data class (or a nested config class like pydantic)? Such a thing would be quite practical for adding group validators as well.

We could assume that grouping is wanted if such a decorator is present? Or an extra boolean paramater therein. Just thinking out loud.

BrianPugh:
If we were to use a decorator, the closes solution would be something like:

```python
@auto_group
@dataclass
class Tester:
    ...
```

where the `auto_group` decorator would essentially return:

```python
Annotated[Tester, Parameter(auto_group=True)]
```

Unfortunately, this isn't a great solution because technically [Annotated objects should not be instantiated](https://typing.readthedocs.io/en/latest/spec/qualifiers.html#annotated):

> An attempt to call Annotated (whether parameterized or not) should be treated as a type error by type checkers

[...]

PR #259, opened 2024-11-23
Parameter.auto_group and utils.default_auto_group (Automatic nested subfield Group creation)

BrianPugh:
This PR allows for the automatic creation of Groups for nested subfields of dataclass-like structures.

@vhdirk can you play around with this branch and see if it works as you'd expect?

Addresses #248


BrianPugh:
@vhdirk bump

BrianPugh:
@vhdirk bump

vhdirk:
Sorry, I've had a busy time recently. I'll take a look at it in the next few days. Thanks a lot already!

BrianPugh:
@vhdirk :smiling

vhdirk:
Ok, so I've finally found a little time to gave it a whirl! Not a lot, so this may come as a bit of a rambling :)

I've been toying around with just the auto_group parameter and the new decorator. At first glance, it seems to work well, though I did encounter some oddities:

- Decorating a (data)class with `@Parameter` seems to work, but if I decorate the class and annotate its use as `Annotated[MyData, Parameter(...)]`, the behaviour struck me as a little counter-intuitive. I would've expected to those to be merged (the decorator has the `auto_group` flag set, the annotation hasn't). I use the same structure multiple times with different names. The  Parameter decorator would allow for setting `common` options, with the annotation as specific overrides?

- `auto_group` seems to immediately apply to all nested structures as well, unrolling them into a (in my case) rather large number of tiny groups. I don't think that's wrong necessarily; if you set auto grouping to the root structure/parameter, it saves having to configure it multiple times for nested structures. In my case though, just setting `group="groupname"` makes it group only that particular dataclass, which is what I needed (reusing the same structure calls for different named groups anyway).

BrianPugh:
other thoughts:
1. Maybe `auto_group` should be an `int` instead of a bool, and it will apply "that many layers deep". This is probably a nice mixture of inheritance and control.

BrianPugh:
closing as stale; we can revisit if the desire arises

Issue #262, opened 2024-11-26
NewType support


m09:
[`NewType`](https://docs.python.org/3/library/typing.html#newtype)'s magic seems to confuse `cyclopts` at the moment. On my Python 3.10, `cyclopts` expects two values when I annotate a command argument with a `str` derived instance of `NewType`. 

It is possible to detect `NewType`s in the following way on my Python 3.10, although methods may vary between versions since `NewType` changed in 3.10 & 3.11:

```python
>>> A = NewType("A", str)
>>> isinstance(A, NewType)
True
>>> A.__supertype__
<class 'str'>
```

BrianPugh:
Thanks for opening the issue! I haven't ever actually seen `NewType` before, but I'm sure we can probably add support. I'll investigate soon, thanks!

m09:
Thanks a lot Brian!

m09:
I had time to test things a bit more and it seems that I proceeded too hastily yesterday during my testing session: I got lists working but only because I used an even number of items during my tests. It seems that an odd number of items in the list will make `cyclopts` fail.

MWE:

```python
from typing import NewType

from cyclopts import App

app = App()
A = NewType("A", str)


@app.default()
def main(items: list[A]) -> None:
    print(items)
```
Output:

```
❯ mwe a                  
╭─ Error ────────────────────────────────────────────────────────────────────────────────────╮
│ Parameter "--items" requires 2 arguments. Only got 1.                                      │
╰────────────────────────────────────────────────────────────────────────────────────────────╯

❯ mwe a b
['a', 'b']

❯ mwe a b c
╭─ Error ────────────────────────────────────────────────────────────────────────────────────╮
│ Parameter "--items" requires 2 arguments. Only got 1.                                      │
╰────────────────────────────────────────────────────────────────────────────────────────────╯

❯ mwe a b c d
['a', 'b', 'c', 'd']

❯ 
```

BrianPugh:
seems like this current issue is specific to some versions of python; investigating.

BrianPugh:
this should now be fixed in v3.1.2!

m09:
Thanks a lot Brian :+1:

PR #263, opened 2024-11-26
Support NewType.

BrianPugh:
@m09 can you play around with this and give feedback? I'm not experienced with `NewType`, so I could use some backup 😄. The actual passed in object never gets cast to the declared new-type, but I don't think that matters since static type checkers  in the parenting project won't be peeking into Cyclopts.

Addresses #262.

### DIFF
--- a/cyclopts/annotations.py
+++ b/cyclopts/annotations.py
@@ -96,6 +96,7 @@ def resolve(type_: Any) -> type:
         type_ = resolve_annotated(type_)
         type_ = resolve_optional(type_)
         type_ = resolve_required(type_)
+        type_ = resolve_new_type(type_)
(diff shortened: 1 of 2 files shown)


m09:
I tried it in a few ways (newtype alone, in a union, in a list, then give it as argument to a function to see if mypy complains) and everything went perfectly fine on my 3.10 (as you suspected I didn't get warnings from mypy even though the object is not cast).

I cannot speak for other python versions but it looks very promising (and what a lightning fast PR!)

Thanks for the great work :)

BrianPugh:
good enough for me! Other python versions are checked by the unit tests. Merging now and this will be available in v3.1.1 momentarily!

Issue #276, opened 2024-12-29
Lazy load subcommands


mrjsj:
Hi - I love the ergonomics of this. It's really a breeze to work with.

I've built a CLI with a lot of nested subapps/subcommands, and since it goes through and registers everything on every invocation, it is a bit slow. 

Is there any way to lazy load subapps and its subcommands? Here's an example.

When I run  `mycli --help` only data-pipeline and environment subapps should be loaded, but not their subapps/commands.
When I run `mycli environment` then only children of the environment subapp should be loaded.

```bash
├── data-pipeline [app]
│   ├── create [command]
│   ├── delete [command]
│   ├── get [command]
│   ├── list [command]
│   └── update [command]
├── environment [app]
│   ├── create [command]
│   ├── delete [command]
│   ├── get [command]
│   ├── list [command]
│   ├── update [command]
│   ├── spark-compute [app]
│   │   ├── get-published-settings [command]
│   │   ├── get-staging-settings [command]
│   │   └── update-staging-settings [command]
│   └── spark-libraries [app]
│       ├── cancel-publish [command]
│       ├── delete-staging-library [command]
│       ├── get-published-libraries [command]
│       ├── get-staging-libraries [command]
│       ├── publish-environment [command]
│       └── upload-staging-library [command]
```

And this is how I do it currently:

```python
from cyclopts import App

from .create import create_command
from .delete import delete_command
from .get import get_command
from .list import list_command
from .spark_compute import spark_compute_app
from .spark_libraries import spark_libraries_app
from .update import update_command

environment_app = App(name="environment", help="[yellow]Commands for environment[/yellow]")
environment_app.command(create_command, name="create")
environment_app.command(delete_command, name="delete")
environment_app.command(get_command, name="get")
environment_app.command(list_command, name="list")
environment_app.command(update_command, name="update")
environment_app.command(spark_compute_app)
environment_app.command(spark_libraries_app)
```

BrianPugh:
Is your project publicly available? I'd like to run a profiler on it. Lazy loading is one solution, but I wonder if something else is slow (and lazy loading would be a premature optimization that wouldn't help too much).

mrjsj:
Yes. 
https://github.com/mrjsj/msfabricutils/tree/feat/new-cli/cli

Would be lovely. Thanks!

mrjsj:
[...]

PR #279, opened 2024-12-29
Various optimizations.

BrianPugh:
After profiling @mrjsj application [msfabricutils](https://github.com/mrjsj/msfabricutils/tree/feat/new-cli/cli) mentioned in #276, I was able to optimize displaying the intial helpscreen from ~570mS to ~450mS on my computer (~26% improvement). On my computer, just importing the main imports (listed below) takes roughly 360mS, suggesting that Cyclopts is taking ~90mS to parse all the commands and to display the help page. If I stub out the actual help-printing, then the resulting program takes ~430mS to execute, suggesting that Cyclopts is taking 20mS to do the help-page parsing and printing, while 70mS doing all the other stuff. Of all of this:

1. 51.5mS are spent in `@app.command`, which is what we could potentially optimize with lazy loading. I'm just super hesitant to do this because lazy-loading can get hairy (unless someone has an elegant solution!).

2. Within `@app.command`, 12.1mS are spent in the recently-optimized `validate_command`. We *could* simply get rid of this; currently its only job is to detect some rare, invalid Cyclopts configurations at application-load time. This is intended so that any cyclopts *programming* error is always detected, rather than only in potentially rare CLI invocations.

3. Within `@app.command`, 26.7mS of time are spent instantiating sub-`App`'s. I imagine most of these are help/version handlers and could be optimized, potentially shaving maybe ~15mS.

---
Main imports:
```
import cyclopts
import msfabricpysdkcore
import json
import requests
import logging
import base64
import rich
import msfabricutils
```

# Changelog
* Removed `validate_command` on register and defer it to execution. The original thinking of wanting to catch programming errors doesn't exactly hold water; it's not like people will program up commands and not test them, which will inherently cause the command to be validated.

### DIFF
--- a/cyclopts/annotations.py
+++ b/cyclopts/annotations.py
@@ -107,9 +107,7 @@ def resolve_optional(type_: Any) -> type:
     """Only resolves Union's of None + one other type (i.e. Optional)."""
     # Python will automatically flatten out nested unions when possible.
     # So we don't need to loop over resolution.
-
-    origin = get_origin(type_)
-    if not is_union(origin):
+    if not is_union(type_):
         return type_
 
     non_none_types = [t for t in get_args(type_) if t is not NoneType]
--- a/cyclopts/annotations.py
+++ b/cyclopts/annotations.py
@@ -6,13 +6,11 @@ from typing import Annotated, Any, Optional, Union, get_args, get_origin
 import attrs
 
 _IS_PYTHON_3_8 = sys.version_info[:2] == (3, 8)
-_union_types = set()
-_union_types.add(Union)
 
 if sys.version_info >= (3, 10):  # pragma: no cover
     from types import UnionType
-
-    _union_types.add(UnionType)
+else:
+    UnionType = object()
 
 if sys.version_info < (3, 11):  # pragma: no cover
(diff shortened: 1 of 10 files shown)


jnk22:
Hi Brian, this is great news!

Just wanted to confirm that this also improved the performance for my app's loading time.

```console
$ hyperfine "app --help"
Benchmark 1: app --help
  Time (mean ± σ):     204.2 ms ±   2.0 ms    [User: 182.5 ms, System: 20.9 ms]
  Range (min … max):   201.2 ms … 207.4 ms    14 runs
```

This is my app's timing before with v3.1.2, this is what I get now with v3.1.3:


```console
$ hyperfine "app --help"
Benchmark 1: app --help
  Time (mean ± σ):     195.4 ms ±   2.4 ms    [User: 179.5 ms, System: 14.9 ms]
  Range (min … max):   191.3 ms … 199.5 ms    15 runs
```

Both mean and minimum times show an improvement of ~5% for me. This is ofc not as much as your 26%, but I only have 5 sub-commands with 3 of them having 2 nested sub-commands, so might be a small app in comparison.

For me, this is really great as I have been measuring the timings via the help display for the last weeks as well.

Issue #282, opened 2025-01-03
Load dict-like parameters from envs


AngellusMortis:
https://github.com/BrianPugh/cyclopts/issues/241 added support for dict-like parameters, i.e. dataclasses and Pydantic classes. It would be nice to be able to load these from an ENV.

```python
from cyclopts import App
from dataclasses import dataclass

app = App()

@dataclass
class User:
   name: str
   age: int

@app.command
def run(
   user: Annotated[User, Parameter(env_var="USER")]
):
   print(user)

app()
```

```bash
$ python scratch.py run --help
Usage: scratch.py run [ARGS] [OPTIONS]

╭─ Parameters ────────────────────────────────────────────────╮
│ *  USER.NAME --user.name  [required]                        │
│ *  USER.AGE --user.age    [required]                        │
╰─────────────────────────────────────────────────────────────╯

# suffix parameters similar to how args work
$ USER_NAME="Bob Smith" USER_AGE=30 python scratch.py run
User(name='Bob Smith', age=30)

# use json to load a dict
$ USER='{"name": "Bob Smith", "age": 30}' python scratch.py run
User(name='Bob Smith', age=30)
```

BrianPugh:
I'd say this is somewhere in between a bug and a feature; regardless I think it would be good to have.

I'm not sure if I'd be immediately able to fix this, but I think you can partially achieve what you want by doing the following:

```python
import cyclopts
from cyclopts import App
from dataclasses import dataclass

app = App(config=cyclopts.config.Env(command=False))

@dataclass
class User:
   name: str
   age: int

@app.command
def run(user: User):
   print(user)

app()
```

```console
$ USER_NAME="Bob Smith" USER_AGE=30 python issue-282.py run
User(name='Bob Smith', age=30)
```

The json-env-string would be useful to parse; I'd have to add that as a new feature.

AngellusMortis:
The config is really useful trick. And yeah, I was thinking it was partial as bug as well considering the ENV still shows up for dict objects, but there is no valid way to provide them.

And I actually really need the being able to parse as JSON part for my use case. It is for basically bootstrapping a config from a k8s manifest, which would be hard to decouple from being JSON.

BrianPugh:
I think I can easily add json-support; let me investigate that.

However, adding support for `USER_NAME` (in your original post, not using `config`) won't be possible; there are ambiguities with environment variable names that don't exist with the classical `--cli.option` names. Another alternative option is:

```python
[...]

PR #285, opened 2025-01-03
Parse json data from token if object is dict-like.

BrianPugh:
If an object is dict/dataclass-like, attempts to parse input data as a json-string.

@AngellusMortis can you play around with this and see if it satisfies your needs? I need to flesh out the rest of the PR and think a little harder if this is the appropriate mechanism, but it naively seems to work.

Addresses #282.

### TODO

- [ ] unit tests
- [ ] docs
- [ ] maybe allow direct assignment to dict-like parameters from the cli (as a json string).


AngellusMortis:
It looks like `data` is getting populated correctly, but the correct values are not being passed to Pydantic correctly (I am using pydantic). It looks like if it was just a basic dataclass, it would work

AngellusMortis:
That commit looks good!

BrianPugh:
@AngellusMortis I just pushed a fix for that; please give it a try! Pydantic is a more-special-than-usual case within cyclopts; as it's the only type that cyclopts practically disables it's own coercion engine for.

BrianPugh:
that was a lightning fast reply!

AngellusMortis:
Watching a show right now. So Github is up and just automatically updating for me. lol.

BrianPugh:
I'd appreciate it if you'd keep playing with it and let me know if you discover any other issues. Otherwise tonight/this weekend I'll be adding tests/docs for this new feature! Thanks!

AngellusMortis:
Find an issue: If you have aliases for your pydantic class, it does not accept the fields from the alias value.

i.e. If you have a config that converts camelCase to snake_case:
```python
class BaseK8sModel(BaseModel):
    """Base model for k8s spec."""

    model_config = ConfigDict(
        alias_generator=to_camel,
        populate_by_name=True,
        from_attributes=True,
    )
```

It cannot load the values from camelCase, only snake_case.

BrianPugh:
I'll look into this, thanks!

AngellusMortis:
@BrianPugh Can you rebase this branch off of `main`? Thanks!

BrianPugh:
rebased! I still want to rethink this a little more; because as of right now it sort of bypasses a lot of Cyclopts mechanisms. It probably works well for most general use-cases (once I investigate the pydantic issue you brought up), but would like it to work for even those advanced use-cases.

BrianPugh:
@AngellusMortis I've updated the implementation to what I believe to be "more proper." I still need to add more unit tests and documentation, but I would appreciate it if you also played with it some! Thanks!

BrianPugh:
I think this feature is functionally complete, I just need to add some docs.

I've also enabled it so that you can do:
```bash
# These 2 get parsed as the same:
$ python my-script.py --my-dataclass.key1 100 --my-dataclass.key2 200
$ python my-script.py --my-dataclass='{"key1": 100, "key2": 200}'
```
Basically if the following conditions are met, the token is parsed as json:
1. The parameter has to be specified as an option; e.g. `--my-dataclass`.
2. The referenced parameter must be dataclass-like (it accepts keys).
3. The referenced parameter **cannot** be union'd with a `str` (so that there's no ambiguity).
4. The first character must be a `{`.

Most ambiguities are pretty much non-issues for dataclass-like objects. However, I have chosen not to extend this functionality to lists quite yet, because `list[str]` is a common annotation, and it directly results in an ambiguous case.

AngellusMortis:
Sorry I did not get a change to test this more as you were working on it, but it seems like now it again cannot load things from aliases. You can load them from the Python name or the alias _lowercased_. 

```python
from typing import Annotated

from pydantic import BaseModel, ConfigDict
from pydantic.alias_generators import to_camel

class BaseK8sModel(BaseModel):
    model_config = ConfigDict(
        alias_generator=to_camel,
        populate_by_name=True,
        from_attributes=True,
    )

class Spec(BaseK8sModel):
    storage_class: str

app = App()

@app.default
def run(spec: Annotated[Spec, Parameter(env_var="SPEC")]) -> None:
    print(spec)

app()
```

```bash
# works
SPEC='{"storage_class": "longhorn"}' python test.py
# works
SPEC='{"storageclass": "longhorn"}' python test.py
# does not work
SPEC='{"storageClass": "longhorn"}' python test.py
```

This is how pydantic dumps the data:

```python
s = Spec(storage_class="longhorn")

assert s.model_dump_json() == '{"storage_class": "longhorn"}'
assert s.model_dump_json(by_alias=True) == '{"storageClass": "longhorn"}'
```

AngellusMortis:
It is also casting `null` into a string before trying to load it. So pydantic is always passed `None` as a string.

BrianPugh:
thanks for the bug report! My schedule is a bit booked, but I'll try and investigate this in the coming week.

BrianPugh:
we can continue conversation in #332

Issue #290, opened 2025-01-08
Nested Commands not showing help correctly / nested meta-apps not running


AngellusMortis:
Definitely a bug or maybe I am just doing things wrong.

I have a nested CLI app structure:

```
app ->
    app_1 ->
        command_1
        command_2
    app_2 ->
        command_1
        command_2
        command_3
```

I also have meta-apps for both the parent app and child apps. So top level app is stuff like logging format, logging level, etc. Then the child apps have more specific global options for all of the grouped child commands. The issue is the help does not display right for all child commands _or_ the meta-apps are not executed.

### Attempt 1

This leads to the only the top-level meta-app being called (`meta_1` is never called):
```python
from cyclopts import App

app_1 = App()

@app_1.meta.default
def meta_1() -> None:
    print("child meta")

app = App()
app.command(app_1)  # <-- difference here

@app.meta.default
def meta() -> None:
    print("parent meta")

if __name__ == "__main__":
    app.meta()
```

### Attempt 2

`meta_1` is called, but only the first two levels of `--help` to work (parent app works, child apps work, but child app commands do not work, it returns the help of the child app instead):
```python
from cyclopts import App

app_1 = App()

@app_1.meta.default
def meta_1() -> None:
    print("child meta")

@app_1.command
def test() -> None:
     pass

app = App()
app.command(app_1.meta)  # <-- difference here

@app.meta.default
def meta() -> None:
    print("parent meta")

if __name__ == "__main__":
    app.meta()
```
i.e. `app --help` works, `app app_1 --help` works, but `app app_1 test --help` displays the help for `app_1`.

BrianPugh:
I never really intended meta apps to be nested this way, but I think it could work. A meta app is very similar to an independently declared `App`, it just gets special treatment relative to it's parent for the `--help` page. I think your Attempt 2 is closest to working. It's not exactly working as expected on my end; i'll try and open up a PR soon and we can continue discussion there.

AngellusMortis:
I am basically just using meta-apps for app wide arguments. Like I mentioned, the top-level app is for stuff like logging level, output format, etc. Then since the current project I am working on is basically a manager for a game server, it is still like hostname/IP, port, RCON port, RCON password, etc. Which that is all data that is shared between the commands in child app, but not the parent app.

BrianPugh:
#291 is in the newly released v3.1.5

PR #291, opened 2025-01-09
Better handling of nested meta apps

BrianPugh:
@AngellusMortis can you play around with this? I think the following script (along with this PR) is kind of what you want:

```python
from typing import Annotated

app_1 = App()

@app_1.meta.default
def meta_1(*tokens: Annotated[str, Parameter(show=False, allow_leading_hyphen=True)]) -> None:
    """This is app1_meta's help."""
    print("child meta")
    app_1(tokens)

@app_1.command
def test() -> None:
    """"app_1 test" help string."""
    print("app_1 test body")

app = App()
app.command(app_1.meta, name="app1")

@app.meta.default
def meta(*tokens: Annotated[str, Parameter(show=False, allow_leading_hyphen=True)]):
    print("parent meta")
    app(tokens)

if __name__ == "__main__":
    app.meta()
```

Comments:
* meta apps **don't** execute for `--help`. This is intentional as frequently expensive operations are performed in a meta app, such as connecting to an external resource (like a database).
* meta apps **do** execute during conventional command execution.
* The `tokens` in the meta signature is [copied from the meta app docs](https://cyclopts.readthedocs.io/en/latest/meta_app.html).

Addresses #290.

### DIFF
--- a/cyclopts/core.py
+++ b/cyclopts/core.py
@@ -136,14 +136,16 @@ def _validate_default_command(x):
     return x
 
 
-def _combined_meta_command_mapping(app: Optional["App"], recurse_meta=True) -> dict[str, "App"]:
+def _combined_meta_command_mapping(
+    app: Optional["App"], recurse_meta=True, recurse_parent_meta=True
+) -> dict[str, "App"]:
     """Return a copied and combined mapping containing app and meta-app commands."""
     if app is None:
         return {}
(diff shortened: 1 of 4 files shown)


AngellusMortis:
Looks like it works!

BrianPugh:
I'll be trying to do a patch release this weekend including this; I want to get some other fixes in with it.

PR #297, opened 2025-01-14
use google for search bar

BrianPugh:
An attempt for using google's more powerful search to replace sphinx's search bar. However, the results are pretty ugly. Anyone have any suggestions?

Docs are built at:
https://cyclopts.readthedocs.io/en/google-search/

### DIFF
--- a/docs/source/_templates/searchbox.html
+++ b/docs/source/_templates/searchbox.html
@@ -1,0 +1,3 @@
+<script async src="https://cse.google.com/cse.js?cx=d7869aba9229a4ca3">
+</script>
+<div class="gcse-search"></div>


AngellusMortis:
Algolia is really popular. It is integrated right into a lot of other static site generators. Looks like you can add it to Sphinx: https://github.com/algolia/sphinx-docsearch

Or as an alternative, something other than Sphinx. There is quite a bit of other popular markdown static site generators for docs. More work, but might be interesting long term if you see them as useful.

Some popular alternatives:

* mkdocs-material: https://github.com/squidfunk/mkdocs-material
* docusaurus: https://github.com/facebook/docusaurus

BrianPugh:
going to close for now; I'm a bit hesitant on Algolia's pricing model; I would prefer something thats "free for all open source."

At some point may move over to a markdown-based document generator, but not at the present.

Issue #238, opened 2024-10-26
A Testing Client


dunossauro:
Hello @BrianPugh,

Thank you so much for the project—it's truly amazing and simplifies many tasks!

I have recently started my journey with Cyclopts, and I've found that testing my CLIs can be quite cumbersome due to the lack of mechanisms that facilitate this.

For example, with a simple app:

```python
from cyclopts import App

app = App()

@app.command
def buzz(n: int):
    print(f"BUZZ: {n}")
```

Testing requires some workarounds, such as using `capsys`:

```python
def test_fizz(capsys):
    app('buzz 5')
    assert capsys.readouterr().out.strip() == 'BUZZ: 5'
```

While I understand it's possible to return the value before printing, as seen in [App Calling & Return Values](https://cyclopts.readthedocs.io/en/latest/app_calling.html), I believe this process could be simplified.

For instance, Typer has a testing client called [CLIRunner](https://typer.tiangolo.com/tutorial/testing/#test-the-app), which is inherited from the [CLIRunner](https://click.palletsprojects.com/en/stable/api/#click.testing.CliRunner) in Click. Additionally, libraries like Cleo also provide a [testing client](https://cleo.readthedocs.io/en/latest/introduction.html#testing-commands).

Implementing something similar in Cyclopts would greatly facilitate application development.

Thank you in advance for considering this suggestion!

BrianPugh:
Thank you for the kind words!

I'm open to discuss the feature of a test runner, but I personally never liked Typer's (and I suppose by extension, Click's) test runner.

Let me try to enumerate the features that Typer's `CliRunner` offers:

1. Checking the application's `exit_code`.
2. Checking `stdout` and `stderr`.
3. Ways of setting environment variables.

I don't think that a test runner really makes any of these operations terser than pytest builtin fixtures. Pytest builtin fixtures also have the benefit of being generic (and not something that someone needs to learn specifically for Cyclopts).

Let's use your example as a basis.

## 1. Checking the application's `exit_code`.
In Typer, you can manage the application's exit code in a few ways:
1. An uncaught exception will result in an `error_code==1`.
2. You can `raise typer.Exit(5)` to result in an `error_code==5`.
3. You can `sys.Exit(5)` that will also result in an `error_code==5`.

[...]

PR #300, opened 2025-01-15
Warn if cyclopts application is invoked with no python arguments within pytest.

BrianPugh:
Displays a warning if a user attempts to invoke a cyclopts application without python arguments while pytest is running. This causes `sys.argv` to be unexpectedly read. I cannot imagine a possible scenario where this would be intentional. This PR only covers `pytest`, but other popular testing frameworks could be added. The next obvious choice would be `unittest`, but I didn't find a definitive way of detecting it was running.

Alternative solution in response to @jayqi https://github.com/BrianPugh/cyclopts/issues/238#issuecomment-2591559815. What do you think of this?

Example output:

```text
=============================== warnings summary ================================
test.py::test_no_args
  /Users/brianpugh/projects/cyclopts2/test.py:64: UserWarning: Cyclopts application invoked without tokens under unit-test framework "pytest". Did you mean "app([])"?
    app(console=None)
```

### DIFF
--- a/cyclopts/core.py
+++ b/cyclopts/core.py
@@ -5,7 +5,8 @@ import traceback
 from collections.abc import Iterable, Iterator
 from contextlib import suppress
 from copy import copy
-from functools import partial
+from enum import Enum
+from functools import lru_cache, partial
 from itertools import chain
 from pathlib import Path
 from typing import (
@@ -898,6 +899,9 @@ class App:
         *,
         console: Optional["Console"] = None,
     ) -> tuple[Callable, inspect.BoundArguments, list[str], dict[str, Any], ArgumentCollection]:
+        if tokens is None:
+            _log_framework_warning(_detect_test_framework())
+
         tokens = normalize_tokens(tokens)
 
         meta_parent = self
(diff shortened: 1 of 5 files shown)


jayqi:
Sorry for the late response. I just upgraded and tested this change. Some thoughts:

- Depending on how much test output there is, the warning may or may not be hard to see, since warnings are printed in a separate section from the failures. 
- My warning message says `UserWarning: Cyclopts application invoked without tokens under unit-test framework "pytest". Did you mean "reprexlite([])"?` where `reprexlite` is the name of my package that I'm testing, even though in the code in my test is `app()`. Not sure how reliable it is to detect the correct symbol name, but the wrong symbol name is maybe more confusing than having it hard-coded to something like "If you're invoking with `app()`, did you mean `app([])`" might be better? 

Anyways, this change certainly doesn't hurt and I think is is an improvement. Thanks for the quick addition. I think it might also be worth calling out as a gotcha in the new [unit testing documentation](https://cyclopts.readthedocs.io/en/latest/cookbook/unit_testing.html) as well.

BrianPugh:
> Depending on how much test output there is, the warning may or may not be hard to see, since warnings are printed in a separate section from the failures.

I agree, but I think it's the best we can do from within Cyclopts. From a philosophical standpoint, I really don't want to add additional test-runner-esque machinery to the codebase.

> My warning message says UserWarning: Cyclopts application invoked without tokens under unit-test framework "pytest". Did you mean "reprexlite([])"?

I'll investigate; thanks for posting your app!

> I think it might also be worth calling out as a gotcha in the new [unit testing documentation](https://cyclopts.readthedocs.io/en/latest/cookbook/unit_testing.html) as well.

Agreed, i'll add something!

PR #305, opened 2025-01-19
`cyclopts.run` WIP

Tinche:


BrianPugh:
Hi @Tinche! Glad you are liking Cyclopts! I like your `cattrs` 😄 .

I'm not fully on board, but I think I could be swayed. Here's me thinking out loud. In these examples I'll be removing the `if __name__ == "__main__"` bit for apples-to-apples comparison for the use-case of short CLI scripts.

The current simplest cyclopts app is 6 lines of code:

```python
from cyclopts import App

app = App()

@app.default
def name_of_script():
    pass

app()
```

For use-case (1), the proposed `run` is 4 lines of code, saving 2 lines of code (33% reduction!). Use-case (2) is also 4 lines of code. I think 4 lines of code is basically the smallest amount you can get, outside of running the script via a `cyclopts` CLI runner.

I'm not worried about cyclopts-code-maintenance for this feature, but what I am worried about is introducing a new (admittedly simple) concept to a Cyclopts user. Currently there is exactly 1 way to define an app and to invoke it. Introducing a second way may cause a Cyclopts-user to wonder "is there more?"

Now, that said, I do want the friction to using cyclopts to be minimized, and I like this PR (or something like this PR) for that reason. One could also argue that "just use `cyclopts.run` for simple scripts" is easier to remember than creating the `App` object and decorating your function with `@app.default` and invoking `app()` at the bottom of the script.

OK, I think I'm on board 😄 .

A thought I was having is that `run` should take `**kwargs` and pass it along to `App`, but maybe we should keep `run` as simple as possible to encourage the current `App()` object syntax for more complicated/configured CLIs.

Additional thoughts: I think using `run` as a decorator should **not** take in a string and should just unconditionally run the function. I think this because developers typically like to look for the `if __name__ == "__main__":` as a hint that the script should be ran from the CLI. While `@run(__name__)` invokes more "what kind of magic is going on" feelings. I also like it when decorators can be used both as `@run` or `@run()`, but in this particular use-case it might result in a user erroneously calling a bare `run()` at the bottom of their script and be left wondering "why are things not running."

Would love to hear your thoughts!

Tinche:
I don't think there are any right or wrong answers here, but let me share my views.

> For use-case (1), the proposed run is 4 lines of code, saving 2 lines of code (33% reduction!).

> I'm not worried about cyclopts-code-maintenance for this feature, but what I am worried about is introducing a new (admittedly simple) concept to a Cyclopts user.

For me it's not just about lines of code, but about the number of concepts. You are correct that this feature would, technically, increase the number of concepts that the users need to be exposed to. I also agree you don't generally want two ways do to something, again due to cognitive load.

I think we should still do it in this case. First of all, having a very streamlined way of accessing the core functionality of a library is very valuable. Not only does it make easier to "sell" the library to non-users (because your early examples can be super focused) but it also reduces cognitive load for novices. Suppose I submit a PR to a colleague at work using the `App` approach - now instead of a single concept, `cyclopts.run`, they need to be aware of `cyclopts.App`, `cyclopts.App.default`, and `cyclopts.App.__call__`. (And classes carry a greater cognitive burden than functions.)

Cattrs is in a similar situation - if you do any serious work with it, you will want to instantiate and customize your own Converter. But we still have `cattrs.structure()` and `cattrs.unstructure()` for use cases that don't require it. I find it works quite well, and there's a clear path to migrating to your own converter if/when you need it.

To me, the word "boilerplate" means someone is making you deal with an API at a level you don't need to be dealing at, introducing unnecessary concepts too early.

> A thought I was having is that run should take **kwargs and pass it along to App, but maybe we should keep run as simple as possible to encourage the current App() object syntax for more complicated/configured CLIs.

Maybe. I think I would like the `run` args to offer some simple configuration in the future. We can start with nothing and go from there?

> Additional thoughts: I think using run as a decorator should not take in a string and should just unconditionally run the function. I think this because developers typically like to look for the if __name__ == "__main__": as a hint that the script should be ran from the CLI. While @run(__name__) invokes more "what kind of magic is going on" feelings. I also like it when decorators can be used both as @run or @run(), but in this particular use-case it might result in a user erroneously calling a bare run() at the bottom of their script and be left wondering "why are things not running."

The decorator use case I think was kind of a long-shot; I don't know the right answers here. Maybe we could make it so that `@run` runs the function automatically, but also offers an optional boolean argument (`run_if=...`, maybe a better name?), so folks can do something like

```
@run(run_if=__name__ == "__main__")
def main():
    ...
```

I think that would make it pretty clear to most Python devs what is going on, and it can be an "advanced" feature mentioned later in the docs.

I would also be good with punting the decorator use case for later, so we can think about it more.

Agreed on the bare `@run` - it should work.

BrianPugh:
Alright, so I think we have some consensus here. I would say the follow action items would leave me happy for an initial release of this feature:

1. Remove the `"__main__"` str check for now; have it only consume a required function. Lets avoid the decorator use-case for now, because it complicates the implementation as well as the docs (e.g. does it return the function, or the result of the call `app.__call__`?).
2. Add docs.

BrianPugh (approved):
LGTM, would you like me to take over and add some reference to it in the docs (beyond the README)?

Tinche:
Sure, sounds good!

eonu:
@BrianPugh I have to say that this was probably the main feature from cyclopts that I personally felt was missing. You're right that in the grand scheme of things it probably doesn't simplify a typical CLI script that much, but as little friction as possible is nice for users.

Really solid work on this library overall - I genuinely have little motivation to continue developing mine, with all the awesome features you're putting out, and the speed at which you're doing it!

And good stuff @Tinche!

Issue #309, opened 2025-01-25
Writing tests for config files


jayqi:
I using the [config files](https://cyclopts.readthedocs.io/en/latest/config_file.html) feature of Cyclopts, and I'm struggling a bit with how to effectively write tests.

My inclination is to write tests where I change my working directory to a temporary directory (via pytest's `tmp_path` and `monkeypatch.chdir`) where I can create config file test inputs. However, I'm finding that these files are not getting picked up when I invoke `app(...)` in my tests. As best as I can tell, the reason this happens is that [`config` property](https://github.com/BrianPugh/cyclopts/blob/863648bcd6e2fa7ccd83a63212c31511311d2371/cyclopts/config/_common.py#L79) on `‎ConfigFromFile` instances is implemented as a cached property, so it has already read config files from previous invocations in other tests and this breaks test isolation. 

Any tips or recipes for a reasonable way to have isolated tests? Manually clearing `_config` on my `ConfigFromFile` instances does not seem like a good approach since it's mucking with the internals of `ConfigFromFile`.

BrianPugh:
Hey @jayqi! Excellent question! I gave this a try for myself, and opened up #310 to get it all working. I hadn't considered this use-case before, and internal-to-cyclopts passing in the absolute path to the cached config reads should fix this issue.

Consider the package:

```python
# demo.py
import cyclopts

app = cyclopts.App(config=cyclopts.config.Json("config.json"))


@app.command
def create(name: str, age: int):
    print(f"{name} is {age} years old.")


if __name__ == "__main__":
    app()
```

The following pytest scripts exercise it:
```python
import pytest
import json
import importlib
import demo

@pytest.fixture(autouse=True)
def reload_demo():
    importlib.reload(demo)

@pytest.fixture(autouse=True)
def chdir_to_tmp_path(tmp_path, monkeypatch):
    monkeypatch.chdir(tmp_path)

@pytest.fixture
def config_path(tmp_path):
    return tmp_path / "config.json"

def test_config_1(config_path, capsys):
    with config_path.open("w") as f:
        json.dump({"create": {"name": "Alice", "age": 30}}, f)

    demo.app("create")
    assert capsys.readouterr().out == "Alice is 30 years old.\n"

def test_config_2(config_path, capsys):
    with config_path.open("w") as f:
        json.dump({"create": {"name": "Bob", "age": 40}}, f)

    demo.app("create")
    assert capsys.readouterr().out == "Bob is 40 years old.\n"
[...]

PR #310, opened 2025-01-26
supply the absolute path to cached config-reads. This is done in case the cwd changes.

BrianPugh:
@jayqi can you give this a try? Will provide more information in #309 .


BrianPugh:
We can add updated pytest docs in another PR; merging this now for a release.

PR #315, opened 2025-01-31
Move config-file-caching responsibility into ConfigFromFile.

BrianPugh:
@jayqi please give this a try!

Now on every app call, the configs will check to see if their cache is stale. It does this by comparing the previously read config's path/mtime/size to the found file's current state. This makes the `importlib.reload` hack during unit testing unnecessary. It should also still provide good performance for anyone who is invoking app multiple times in the same session with the same actual config-file.

Overall, this seems like a much more principled, clean approach than the previous caching implementation.

Addresses #309

### DIFF
--- a/cyclopts/config/_common.py
+++ b/cyclopts/config/_common.py
@@ -46,6 +46,30 @@ def _meta_arguments(apps: list["App"]) -> "ArgumentCollection":
     return argument_collection
 
 
+class CacheKey:
+    """Abstraction to quickly check if a file needs to be read again.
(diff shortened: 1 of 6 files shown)


jayqi:
Just installed this branch and removed the `importlib.reload`s in my tests, and everything worked like a charm. ✨

BrianPugh:
good to hear! I'll just have to update the docs about the `monkeypatch.chdir` trick, and then this should be good to go.

PR #323, opened 2025-02-05
run: fix type signature for coroutines

Tinche:
Fixes the type of `cyclopts.run` when the given callable is a coroutine function.

Without this, the wrong return type will be inferred, and Mypy will complain:

```python
error: Value of type "Coroutine[Any, Any, int]" must be used  [unused-coroutine]
note: Are you missing an await?
```

Let me know if you think this requires a changelog entry.

### DIFF
--- a/cyclopts/core.py
+++ b/cyclopts/core.py
@@ -2,7 +2,7 @@ import inspect
 import os
 import sys
 import traceback
-from collections.abc import Iterable, Iterator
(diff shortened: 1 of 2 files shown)


BrianPugh (approved):

BrianPugh:
looks good to me! will merge/release shortly.

BrianPugh:
released in v3.5.1!

Tinche:
Super quick turnaround, thanks!

PR #340, opened 2025-03-05
Docs: Exit non-zero on exception when pprint with rich

thejcannon:
(this one bothered me)


docs/source/cookbook/rich_formatted_exceptions.rst
```diff
@@ -59,6 +60,7 @@ To create a more visually appealing and informative traceback, you can use the `
             app()
         except Exception:
             console.print_exception()
+            sys.exit(1)
```
thejcannon, line 63: Alternatively `raise SystemExit(1) from None`

BrianPugh:
good catch! thanks!

PR #343, opened 2025-03-06
Type fix

kiyoon:
My basedpyright type checker was showing error

### DIFF
--- a/cyclopts/_convert.py
+++ b/cyclopts/_convert.py
@@ -160,7 +160,13 @@ def _convert(
         type_, cparam = Parameter.from_annotation(type_)
         if cparam._converter:
             converter_needs_token = True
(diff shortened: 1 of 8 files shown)


kiyoon:
image

So the problem is that the basedpyright type checker emits this error. It thinks that the default parameter for the converter is None but the type is clearly not allowing None. What is the best way to fix this? I think my fix breaks the app.

BrianPugh:
should be fixed now; i also fixed some other basedpyright errors, but I don't quite have the patience to deal with them all (e.g. changing `Callable` -> `Callable[..., Any]`). I'll cut a release soon, thanks!

kiyoon:
I use basedpyright with standard mode so I think it doesn't have to be perfect to have it fixed on my side. Thank you!!

Issue #352, opened 2025-03-10
Cannot parse dataclass parameter when importing `annotations` (v3.9.2)


beskep:
main.py
```python
from __future__ import annotations

from dataclasses import dataclass

import cyclopts


@dataclass
class Movie:
    title: str
    year: int


app = cyclopts.App(name="movie-maintainer")


@app.default
def add(movie: Movie):
    print(f"Adding movie: {movie}")


if __name__ == "__main__":
    app()
```

---

```console
$ uv add cyclopts==3.9.1
Resolved 11 packages in 15ms
$ uv sync
Resolved 11 packages in 0.71ms
Uninstalled 1 package in 5ms
Installed 1 package in 26ms
 - cyclopts==3.9.1
 + cyclopts==3.9.2
$ uv run .\main.py BladeRunner 1982
Adding movie: Movie(title='BladeRunner', year=1982)
$ uv add cyclopts==3.9.2
Resolved 11 packages in 16ms
$ uv sync
Resolved 11 packages in 0.71ms
Uninstalled 1 package in 5ms
Installed 1 package in 26ms
 - cyclopts==3.9.1
 + cyclopts==3.9.2
$ uv run .\main.py BladeRunner 1982
╭─ Error ──────────────────────────────────────────────────────────────────────────────────╮
│ Invalid value for "MOVIE.TITLE": unable to convert "BladeRunner" into str.               │
╰──────────────────────────────────────────────────────────────────────────────────────────╯
```

No problem without `from __future__ import annotations`.

beskep:
when importing `annotations`, `cyclopts.field_info._dataclass_field_infos` gives `annotation='str'` for `title` field, not `annotation=<class 'str'>`

beskep:
msgspec's [get_dataclass_info()](https://github.com/jcrist/msgspec/blob/dd965dce22e5278d4935bea923441ecde31b5325/msgspec/_utils.py#L233) and [get_class_annotations()](https://github.com/jcrist/msgspec/blob/dd965dce22e5278d4935bea923441ecde31b5325/msgspec/_utils.py#L123) might be helpful.

BrianPugh:
I'm able to reproduce, will investigate shortly, thanks!

BrianPugh:
hmmm, the fix for this (I think) is going to be pretty nontrivial; will have to think about it more.

BrianPugh:
[...]

PR #355, opened 2025-03-12
Partially fix `from __future__ import annotations`

BrianPugh:
Hopefully this solves the problem "good enough." This still doesn't work whenever the annotated type hint is defined in a local scope, but I think that is generally relatively rare given Cyclopts use-case. From my initial attempts, a solution that resolves the surrounding closure is really complicated. I think this PR mostly solves the problem in a pragmatic sense, without adding a ton of code to maintain.

@beskep can you give this a try? What are your thoughts on it?

Addresses #352.

### DIFF
--- a/cyclopts/argument.py
+++ b/cyclopts/argument.py
@@ -2,7 +2,7 @@ import inspect
 import itertools
 from contextlib import suppress
 from functools import partial
-from typing import Any, Callable, Optional, Sequence, Union, get_args, get_origin
+from typing import Any, Callable, Optional, Sequence, Union, get_args, get_origin, get_type_hints
 
 from attrs import define, field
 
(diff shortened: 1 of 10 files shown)


beskep:
I usually use the dataclass argument, so I'm happy with fix.
But NamedTuple (`annotation=ForwardRef('str')`) and attrs (`annotation='str'`) parameters still have the issue.

BrianPugh:
I added tests for attrs and typeddict, can you share the code that leads to an error? Thanks!

beskep:
NamedTuple still has the issue. The other's are fine.

```python
from __future__ import annotations

from typing import NamedTuple, TypedDict

import attrs
from pydantic import BaseModel

import cyclopts


@dataclass
class _DataClass:
    arg: str


class _NamedTuple(NamedTuple):
    arg: str


class _TypedDict(TypedDict):
    arg: str


@attrs.define
class _Attrs:
    arg: str


class _Pydantic(BaseModel):
    arg: str


app = cyclopts.App()


@app.default
def main(dc: _DataClass, nt: _NamedTuple, td: _TypedDict, at: _Attrs, pd: _Pydantic):
    print(dc, nt, td, at, pd)


if __name__ == "__main__":
    app("--dc.arg dc --nt.arg nt --at.arg at --pd.arg pd --td.arg td")
```

╭─ Error ───────────────────────────────────────────────────────────╮
│ Invalid value for --nt.arg: 'ForwardRef' object is not callable   │
╰───────────────────────────────────────────────────────────────────╯

BrianPugh:
fixed! Thanks for testing! I'll merge this in now and cut a release

Issue #361, opened 2025-03-13
ExistingFile doesn't trigger on parameter with default


Tinche:
Hello!

I have a parameter of type `ExistingFile` with a default value. But if the value is not provided (the default is used), the check doesn't fire, and instead of a helpful Cyclopts error I get a stack trace later in the file.

Here's a simple reproducer:

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
#   "cyclopts",
# ]
# ///
from pathlib import Path

from cyclopts import run
from cyclopts.types import ExistingFile


async def main(input_filename: ExistingFile = Path("filename.txt")) -> None:
    print(input_filename)


if __name__ == "__main__":
    run(main)
```

BrianPugh:
hmmm, this is certainly unintuitive/unexpected for a user. Currently Cyclopts doesn't touch/validate any parameters that are not provided via the cli. Generally this is what people would want/expect, but in this situation it definitely feels wrong.

If we take `attrs` as the template of "how it should be done", validators and converters are always triggered.

In the Cyclopts context, converters _usually_ only make sense when going from a CLI token, so I think it's correct that we skip them and take the default value as is. This is also probably what people would intuitively expect (default value comes in unmangled).

I think we should probably *always* run validators, so that your example would behave more as expected. I'd have to think/explore more about how this would interact with something like `file: Optional[ExistingFile] = None`.

What are your thoughts?

BrianPugh:
sounds good, I can tackle this in a few days!

BrianPugh:
Addressed in v3.10.0!

PR #364, opened 2025-03-14
Always apply validator to default signature values.

BrianPugh:
@Tinche I think this sufficiently addresses #361. A downside of this is that there are more complicated interactions with `Union` type hints, but I think you would only run into these edge cases if you're intentionally trying to break Cyclopts. For example, if you were to Union 2 Annotated types with different validators. I did make sure that common use-cases like `Optional[ExistingPath]` work. I might play around with this more before merging to explore more Union-interactions.


Tinche:
Amazing, thanks!

PR #384, opened 2025-03-26
Ensure App is permitting subclasses

thejcannon:
(I'd love to make a test for this but was a smidge lazy, feel free to push back)

Simple change to add some flexibility where clients can subclass `App`, and use `.command()` on a subclass'd App as well.


BrianPugh:
looks good to me! Released in v3.11.1

PR #391, opened 2025-04-03
add datetime and timedelta support.

BrianPugh:
@paw-lu can you play around with this and see if everything works as you would expect? thanks!


paw-lu:
you're really fast 😅 .

yeah messed around with this and it's great. tbh exactly what i had in mind

BrianPugh:
released in v3.12.0!

kiyoon:
This dude is literally going to make something like Pydantic. I love the package and the features being added. Thanks a lot!

Issue #393, opened 2025-04-04
Surprising - multi-line summary line in a command docstring results in empty help for that command.


RandallPittmanOrSt:
First off, cyclopts is awesome, thank you for all the hard work.

I just started using cyclopts and I couldn't figure out why I wasn't getting any help line for a command. Turns out it was because it had a multi-line summary line, which (TIL) is contrary to PEP-0257, and the functionality you've implemented in #82, due to #74.

Editorial: I'm not sure how many people are aware of that bit of PEP-0257, and the language "it is important" in the PEP implies "should" rather than "shall", IMHO. I get this is more of a complaint for the upstream docstring_parser, but what I'm mainly trying to get at was that I found it surprising not to have any help text for a command and it was challenging to track this down. I would suggest either (a) changing help.py to raise an error if a multi-line summary line is detected, or (b) adding a warning in the docs that if the summary line is multiple lines, no help text will appear.

BrianPugh:
Perhaps I might be re-stating what you are saying, but what do we think about raising an exception if a long-description exists, but the short-description is blank? Because that would probably an unintentional developer error.

RandallPittmanOrSt:
Almost. I think an exception could be raised if a `long_ _description` exists and the `short_description` from `docstring_parser` does not match your `short` docstring (`inspect.cleandoc(doc).split("\n\n")`).

Below is an example script that shows how some different docstrings are currently handled by cyclopts (Python 3.12.6, cyclopts 3.12, docstring_parser 0.16). I think it would be good to raise an error for `foo()`'s docstring, because the "short" summary line as you parse it in `help.py` (splitting on double-line-break) is not the same as what docstring_parser puts out (splitting on the first line break).

```python
import textwrap

import docstring_parser
from cyclopts import App

app = App()


@app.command
def foo() -> None:
    """This is a
    multi-line summary line.

    (in violation of PEP-0257).
    """
    pass


@app.command
def bar():
    """This is a single-line summary line.

    Extra stuff...
    possibly continuing more more lines...

    including multiple-line breaks.
    """


@app.command
def baz():
    """

    This summary line is way down here.
    """


if __name__ == "__main__":
    print("----CLYCLOPTS HELP----")
    app(["--help"])
    print("\n----docstring_parser.Docstring short and long descriptions----")
    for fn_name in ["foo", "bar", "baz"]:
        fn = locals()[fn_name]
        ds = docstring_parser.parse(fn.__doc__)
[...]

PR #401, opened 2025-04-08
Handle the first block of docstring text as a potential multi-line short description.

BrianPugh:
Addresses #393. @ RandallPittmanOrSt what do you think of this solution? If we like it, i'll fix tests.

### DIFF
--- a/cyclopts/help.py
+++ b/cyclopts/help.py
@@ -43,14 +43,22 @@ def docstring_parse(doc: str, format: str):
     import docstring_parser
 
     cleaned_doc = inspect.cleandoc(doc)
-    if format != "plaintext":
-        if len(short_description_and_remainder := cleaned_doc.split("\n\n", 1)) == 2:
-            # Place multi-line summary into a single line.
-            # This kind of goes against PEP-0257, but any reasonable CLI command will
-            # have either no description, or it will have both a short and long description.
-            short = short_description_and_remainder[0].replace("\n", " ")
-            cleaned_doc = short + "\n\n" + short_description_and_remainder[1]
+    short_description_and_maybe_remainder = cleaned_doc.split("\n\n", 1)
+
+    # Place multi-line summary into a single line.
+    # This kind of goes against PEP-0257, but any reasonable CLI command will
+    # have either no description, or it will have both a short and long description.
+    short = short_description_and_maybe_remainder[0].replace("\n", " ")
+    if len(short_description_and_maybe_remainder) == 1:
+        cleaned_doc = short
+    else:
+        cleaned_doc = short + "\n\n" + short_description_and_maybe_remainder[1]
+
     res = docstring_parser.parse(cleaned_doc)
+
+    # Ensure a short description exists if there's a long description
+    assert not res.long_description or res.short_description
+
     return res
 
 
--- a/cyclopts/help.py
+++ b/cyclopts/help.py
@@ -11,6 +11,7 @@ from typing import (
     Callable,
     Literal,
     Optional,
+    Sequence,
     get_args,
     get_origin,
 )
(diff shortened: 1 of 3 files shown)


tuukkamustonen:
Works for https://github.com/BrianPugh/cyclopts/issues/402.

For https://github.com/BrianPugh/cyclopts/issues/393, it also works for all cases, except when `help_format="plaintext"` is given - then `foo` shows as:

```
│ foo        This is a                                                                                                                                                                                                                                                                                                                                                                                                                   ```

```py
@app.command
def foo() -> None:
    """This is a
    multi-line summary line.
    
    (in violation of PEP-0257).
    """
    pass
```

BrianPugh:
@tuukkamustonen so that was intentional (I had to add extra code to make it that way), but perhaps also a bit misguided. My idea behind **not** unwrapping the multi-line short-description for "plaintext" docs is that the "plaintext" formatting doesn't perform any unwrapping or text reflowing for the rest of the docstring. It (more or less) is provided exactly as-is.

Now, that said, perhaps it might be more useful to remove the special-case of not interpreting the multi-line docstring for "plaintext". What are your thoughts?

tuukkamustonen:
> Now, that said, perhaps it might be more useful to remove the special-case of not interpreting the multi-line docstring for "plaintext". What are your thoughts?

It's probably a more complicated topic than what it gives by the looks 😄 Tried to Google it, but I don't see many mentions of how to work with multi-line _summary line_.

Anyway, aren't all of those `help_format` options, ie. Markdown, reStructuredText, Rich, styles matters of _formatting_? Ie. colors, **bold**, *cursive*, etc.

For the _structure_ of the docstring on the other hand, should stick to PEP 257 and do it the same way. For the first line vs rest of the content, at least.

The bottom of [this section](https://peps.python.org/pep-0257/#handling-docstring-indentation) shows that the first line can be wrapped into two/multiple lines. I think that should be followed also for `plaintext`.

Plaintext comes without formatting and doesn't allow special constructs (like describing arguments so that they would be interpreted by the parser), but it still must obey to the summary line vs rest of the content separation which sits on top of everything else?

tuukkamustonen:
Okay, scratch that answer. If PEP 257 really requires that _multi-paragraph_ docstrings can't have _multi-line summary line_, then I agree with https://github.com/BrianPugh/cyclopts/issues/393#issuecomment-2784152080 that it should maybe raise an error. The explanation same as above, it still must obey PEP 257 even if it comes without all formatting / syntactic goodies.

BrianPugh:
> Anyway, aren't all of those help_format options, ie. Markdown, reStructuredText, Rich, styles matters of formatting?

of course! But this also includes "how do I interpret newlines?" For example, in markdown, in majority of cases, a single newline simply gets skipped over (not to be confused with 2 newlines "\n\n").

> For the structure of the docstring on the other hand, should stick to PEP 257 and do it the same way.

And we mostly are, but evidently from this and other discussions, people have different opinions on how to handle longer, multi-line short descriptions 😅 . The strictly-to-PEP257 suggestion is that they don't exist, but this also can take some developers by surprise. Supporting it is also  fairly trivial.

> Okay, scratch that answer. If PEP 257 really requires that multi-paragraph docstrings can't have multi-line summary line, then I agree with https://github.com/BrianPugh/cyclopts/issues/393#issuecomment-2784152080 that it should maybe raise an error

My main issue with this is that it can be so obvious/reasonable the intent of a developer to type up a multi-line summary. It's also trivial to support on Cyclopts' side. However, I also want to make sure that such a feature doesn't have unexpected hard-to-debug consequence. I would consider #393 as unexpected and hard-to-debug. With this PR, I think it would hopefully be more obvious how the docstring was being interpreted by Cyclopts that a developer would immediately know how to modify their docstring.

Now, specifically for "plaintext", it would be better to remove the new additional multi-line-short-description logic, because nowhere else in "plaintext" do we do any newline-processing.

tuukkamustonen:
So it's not black and white, there's apparently no right or wrong answer, and you have a justified stance on the topic. I would suggest you just go with it (unless @RandallPittmanOrSt chimes in).

Supporting multi-line summary lines is fine, as you say, because they don't break anything. Linters (Ruff) mostly prevent people from doing those, but if someone does, cyclopts supports them. I don't see harm.

Regarding `plaintext`, if that's how you define it, ie. absolutely no interpretation / extra support, then that's how it is. Just mention that in the docs, and off we go 😄

RandallPittmanOrSt:
I apologize for not responding sooner about all this. I don't have any opinion on `plaintext`, but yeah I think this impl. looks great. Do we need to add a test for multi-line summary though?

Issue #403, opened 2025-04-10
False positive for "application invoked without tokens under unit-test framework" (kinda)


JacobHayes:
As part of my tests, I execute code that shells out to some `cyclopts` scripts. The scripts use the normal `if __name__ == "__main__": app()` logic.

This triggers the `application invoked without tokens under unit-test framework` warning. While _technically_ correct 😅, it's not actually flagging an actual issue here: the `app()` call is not called _directly_ by pytest, but by the normal script execution.

I know I can suppress this, but just figured I'd flag in case it's easy enough to adjust the detection. Perhaps instead of checking for the `PYTEST_VERSION` env var, there's an in-process value that we could detect (though that might require attempting to import `pytest` 😞)?

---

Here's a minimal repro:

`myscript` (`chmod +x ...`):
```python
#!/usr/bin/env python3
from cyclopts import App

app = App()


@app.command()
def check():
    pass


if __name__ == "__main__":
    app()
```

`test_cli.py`:
```python
def test_cyclopt_cli():
    import subprocess

    subprocess.check_call(["myscript", "check"])
```

BrianPugh:
hmmm, we could maybe add some mechanism, but is there any reason to use `subprocess` rather than just directly invoking the `app` object?  i.e:

```python
app(["check"])
```

JacobHayes:
In the real code not called directly in my test code, but deeper in code I'm trying to test. The script uses its own virtual environment with different dependencies, so I cannot import it and basically want to treat it like an opaque executable.

BrianPugh:
gotcha, makes sense, let me see what i can do!

PR #404, opened 2025-04-10
Check for "pytest" in `sys.modules` when checking for PYTEST env vars.

BrianPugh:
Addresses #403. @JacobHayes what do you think of this solution?

### DIFF
--- a/cyclopts/core.py
+++ b/cyclopts/core.py
@@ -1478,7 +1478,7 @@ def _detect_test_framework() -> TestFramework:
         Name of the testing framework. Returns an empty string if not testing
         framework discovered.
     """
-    if os.environ.get("PYTEST_VERSION") is not None:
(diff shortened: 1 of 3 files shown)


BrianPugh:
it's not a perfect solution, but it should resolve your situation.

JacobHayes (approved):
Ah, good idea - thanks for the quick fix!

BrianPugh:
so unfortunately, this check doesn't work if code-cov is enabled, because [code-cov injects a bunch of stuff into subprocess](https://pytest-cov.readthedocs.io/en/latest/subprocess-support.html) environment (which is why my tests are failing).

BrianPugh:
so I guess this is better than the current situation, but will still raise an erroneous warning if code-cov is enabled. For now, I think that's as good as we're going to get.

JacobHayes:
Thanks for the fix here!

Issue #408, opened 2025-04-16
Pydantic Dependency Issue when not using pydantic


domWalters:
I have a `cyclopts` application installed in a `venv` with an old version of `pydantic` in it (1.9.1).

This is somewhat unavoidable for me at the moment; I can't isolate the `cyclopts` app into a different `venv`.

I am not using `pydantic` in my application, but `cyclopts` sees the installation and tries to setup some functions for it at runtime.

This causes an error here: https://github.com/BrianPugh/cyclopts/blob/main/cyclopts/argument.py#L1138

```console
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/cyclopts/argument.py", line 1138, in validate_pydantic
    pydantic.TypeAdapter(hint).validate_python(val)
AttributeError: module 'pydantic' has no attribute 'TypeAdapter'
```

I don't think `TypeAdapter` is in the version of `pydantic` on this machine.

This then causes another error:

```console
During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/share/pynq-venv/bin/aurora-test", line 8, in <module>
    sys.exit(app())
  File "/usr/local/lib/python3.10/dist-packages/cyclopts/core.py", line 1181, in __call__
    command, bound, _ = self.parse_args(
  File "/usr/local/lib/python3.10/dist-packages/cyclopts/core.py", line 1101, in parse_args
    command, bound, unused_tokens, ignored, argument_collection = self._parse_known_args(
  File "/usr/local/lib/python3.10/dist-packages/cyclopts/core.py", line 991, in _parse_known_args
    bound, unused_tokens = create_bound_arguments(
  File "/usr/local/lib/python3.10/dist-packages/cyclopts/bind.py", line 351, in create_bound_arguments
    argument_collection._convert()
  File "/usr/local/lib/python3.10/dist-packages/cyclopts/argument.py", line 197, in _convert
    argument.convert_and_validate()
  File "/usr/local/lib/python3.10/dist-packages/cyclopts/argument.py", line 1182, in convert_and_validate
    self.validate(self.field_info.default)
  File "/usr/local/lib/python3.10/dist-packages/cyclopts/argument.py", line 1161, in validate
    validate_pydantic(self.field_info.annotation, value)
  File "/usr/local/lib/python3.10/dist-packages/cyclopts/argument.py", line 1141, in validate_pydantic
    except pydantic.PydanticUserError:
[...]

PR #409, opened 2025-04-16
don't use pydantic features if pydantic v1 is installed. Addresses #408

BrianPugh:
@domWalters can you confirm that this works for you? Thanks!


domWalters:
> @domWalters can you confirm that this works for you? Thanks!

Yes, replacing the hack I mentioned [here](https://github.com/BrianPugh/cyclopts/issues/408#issuecomment-2809931280) with the diff in this PR also results in avoiding the error.

---

Congrats on winning the award for worlds fastest bug fix!

Issue #407, opened 2025-04-16
Question: Is there a way to textually format the help default for a parameter?


domWalters:
Hi,

I recently discovered your library after getting frustrated with how hard it can be to get `typer` to do what I wanted, and it has been a joy to work with so far.

---

My question relates to the `--help` output.

I have a help window that looks like this:

```console
╭─ Parameters ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --address               [default: 2684358656]                                                                                                                        │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

I'd like the `--address` default to be formatted in hexadecimal, as this is more helpful.

Is there a way to do this without changing the default to be a `str`?

I've looked through the current documentation, and have been unable to workout how to do this if it is possible.

---

Thanks!

BrianPugh:
There isn't currently a way of achieving what you want, short of changing the default to being a str.

As someone who works with a lot of hex, I totally understand the want for this feature.

```python
def _transform(value: Any) -> str:
    return f"0x{value:08X}"

def my_command(
    address: Annotated[int, Parameter(help_default_transform=_transform)] = 0xa0001000,
):
    pass
```

This matches the naming convention of `Parameter.name_transform`.  This could probably also support [template strings](https://peps.python.org/pep-0750/) in the future.

What I don't like about this solution is how clunky it is. We could do something like `Parameter(hex=True)`, but I don't like that either because it puts type-specific configuration into `Parameter`. Can you think of any alternative syntax/solution?

domWalters:
Something like `default_transform` was how I expected it to be done, if the feature was already supported, as I had noticed `name_transform` in the docs.

---

I agree that the `Parameter(hex=True)` approach doesn't sound like a good idea.

---

The `cyclopts.types` idea sounds nice.

Although worth noting, from a quick glance at the source of that file, this would be the first instance of using that module to store "formatting" stuff as far as I can tell. Everything else in there is either a validator or an input converter.

BrianPugh:
> this would be the first instance of using that module to store "output formatting" stuff as far as I can tell

I think that's fine, its goal is to remove `Annotated[..., Parameter(...)]` clunkiness from common use-cases, and I would say this is a fairly common use-case.

I'm going to be busy for ~1week. I might not be able to work on this until then, so we can sleep on it for a little to see if we'd still be happy with it.

[...]

PR #410, opened 2025-04-17
Allow Parameter.show_default to be a callable. Add `HexUInt*` types.

BrianPugh:
Previously, `Parameter.show_default` had type `None | bool`. This PR extends the functionality, where if the provided value is a callable, we will invoke it to translate the default value into a string.

Addresses #407. @domwalters what do you think of this solution? Since this is a relatively specific use-case, I'm trying to avoid adding additional `Parameter` attributes.

### DIFF
--- a/cyclopts/argument.py
+++ b/cyclopts/argument.py
@@ -713,7 +713,7 @@ class Argument:
         return any(dict in (arg, get_origin(arg)) for arg in args)
 
     @property
-    def show_default(self) -> bool:
+    def show_default(self) -> Union[bool, Callable[[Any], str]]:
         """Show the default value on the help page."""
         if self.parameter.show_default is None:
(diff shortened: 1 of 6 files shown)


domWalters:
@BrianPugh this looks great!

I'm off work now until Thursday (24th), so won't be able to test on my application until then.

BrianPugh:
@domWalters have you had an opportunity to try this out? I'm inclined to merge it in this weekend for a release.

domWalters:
> @domWalters have you had an opportunity to try this out? I'm inclined to merge it in this weekend for a release.

Apologies for the late reply.

Just updated to `v3.14.0`, changed the `UInt32` parameter I had to `HexUInt32` and it worked as I expected it to.

Thanks!

PR #426, opened 2025-05-02
Bugfix/dataclass `Parameter.negative` inheritance

### LINKED ISSUE #425: Cannot override negative default for dataclass fields
This does not work:

```python
from pathlib import Path
from pprint import pprint

app = App(default_parameter=Parameter(negative=()))

@Parameter(name="*", negative=())
@dataclass
class Common:
    debug: bool = False
    verbose: bool = False
    out: Path = "out/"

@app.command()
def display(common: Common | None = None):
    if common is None:
        common = CommonArgs()
    pprint(common)

@app.command()
def process(files: list[str], common: Common | None = None):
    if common is None:
        common = CommonArgs()
    pprint(common)

if __name__ == "__main__":
    app()
```

output:

```bash
$ python test.py process --help
Usage: test.py process [ARGS] [OPTIONS]

╭─ Parameters ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ *  FILES --files                   [required]                                                                                    │
│    DEBUG --debug --no-debug        [default: False]                                                                              │
│    VERBOSE --verbose --no-verbose  [default: False]                                                                              │
│    OUT --out                       [default: out/]                                                                               │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

Neither the App-level declaration, nor the class-level, seem to modify the dataclass (but *is* working for the files argument, at least).

At the individual dataclass field level, I'm unable to find any syntax that would allow this kind of override, either (and I'd prefer to keep this code DRY).

Am I missing something obvious?

BrianPugh:
Should fix #425, @wohali can you give this a try? The fix was super simple, I'm not sure what my original intent of blocking specifically `Parameter.negative` from the parameter-resolution-hierarchy. Thanks!

### DIFF
--- a/cyclopts/argument.py
+++ b/cyclopts/argument.py
@@ -54,7 +54,6 @@ _PARAMETER_SUBKEY_BLOCKER = Parameter(
     name=None,
     converter=None,  # pyright: ignore
     validator=None,
-    negative=None,
     accepts_keys=None,
     consume_multiple=None,
     env_var=None,
--- a/tests/test_help.py
+++ b/tests/test_help.py
@@ -1,4 +1,5 @@
 import sys
+from dataclasses import dataclass
 from enum import Enum
 from functools import partial
 from textwrap import dedent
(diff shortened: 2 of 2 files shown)


wohali:
@BrianPugh 👍🏼 good to go

PR #431, opened 2025-05-07
Add (optional) trio support

aneeshusa:


aneeshusa:
Hey @BrianPugh, still using cyclopts and loving it! This is the one feature I need to use this widely. LMK what you think of the approach and how you want me to handle the coverage failure - could add an exclude if we don't want to include trio in the test suite, or I can add trio as an optional dep and add some tests covering it.

cyclopts/core.py
```diff
@@ -23,6 +23,7 @@
 )
 
 from attrs import define, field
+from typing_extensions import assert_never
```
BrianPugh: We only have `typing_extensions` listed as a dependency for python <3.11. So can we change this to:

```python
if sys.version_info < (3, 11):  # pragma: no cover
    from typing_extensions import assert_never
else:  # pragma: no cover
    from typing import assert_never
```

cyclopts/core.py
```diff
@@ -1211,9 +1213,21 @@ def __call__(
...
+            # We don't use anyio to avoid the dependency for non-async users.
+            # anyio can auto-select the backend when you're already in an async context,
+            # but here we're creating the top-level event loop & must select ourselves.
+            if backend == "asyncio":
+                import asyncio
 
-            return asyncio.run(command(*bound.args, **bound.kwargs))
+                return asyncio.run(command(*bound.args, **bound.kwargs))
+            elif backend == "trio":
+                import functools
```
BrianPugh: `partial` is already imported at the top of this file, so lets just directly use that.

BrianPugh:
Oh, and can you also please update the docstring? Thanks!

aneeshusa:
Appreciate the quick review! Should all be addressed now so ready for another look when you get a chance.

BrianPugh:
Released in v3.15.0. Thanks for the contribution!

Issue #432, opened 2025-05-08
Whole-User class validators are broken


wohali:
I'm trying to perform validation on a user class, similar to how a group validator might work, but all of the necessary values are inside of the class itself. Seemed logical to me to use a class validator. But that's causing a traceback:

issue.py:

```python
from cyclopts import App, Parameter
from dataclasses import dataclass

app = App(name="issue")

def v(type_, value):
    raise ValueError("buh")

@Parameter(validator=v)
@dataclass
class Movie:
   title: str
   year: int

@app.command
def add(movie: Movie):
   print(f"Adding movie: {movie}")

app()
```

Execution:
```python
$ python ./issue.py add "Mad Max: Fury Road" 2015
Traceback (most recent call last):
  File "/tmp/cyclopts-issue/.venv/lib/python3.11/site-packages/cyclopts/argument.py", line 1164, in validate
    validator(self.hint, value)
  File "/tmp/cyclopts-issue/./issue.py", line 7, in v
    raise ValueError("buh")
ValueError: buh

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/tmp/cyclopts-issue/.venv/lib/python3.11/site-packages/cyclopts/core.py", line 1123, in parse_args
    command, bound, unused_tokens, ignored, argument_collection = self._parse_known_args(
                                                                  ^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/cyclopts-issue/.venv/lib/python3.11/site-packages/cyclopts/core.py", line 1013, in _parse_known_args
    bound, unused_tokens = create_bound_arguments(
                           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/cyclopts-issue/.venv/lib/python3.11/site-packages/cyclopts/bind.py", line 351, in create_bound_arguments
    argument_collection._convert()
  File "/tmp/cyclopts-issue/.venv/lib/python3.11/site-packages/cyclopts/argument.py", line 196, in _convert
    argument.convert_and_validate()
  File "/tmp/cyclopts-issue/.venv/lib/python3.11/site-packages/cyclopts/argument.py", line 1184, in convert_and_validate
    self.validate(val)
  File "/tmp/cyclopts-issue/.venv/lib/python3.11/site-packages/cyclopts/argument.py", line 1167, in validate
    raise ValidationError(exception_message=e.args[0] if e.args else "", argument=self) from e
cyclopts.exceptions.ValidationError: <exception str() failed>

[...]

PR #433, opened 2025-05-08
Bugfix: Parsing message for class validator

BrianPugh:
@wohali can you give this a try? thanks!

Addresses #432 .

### DIFF
--- a/cyclopts/exceptions.py
+++ b/cyclopts/exceptions.py
@@ -141,10 +141,14 @@ class ValidationError(CycloptsError):
         message = ""
         if self.argument:
             value = self.argument.value if self.value is cyclopts.utils.UNSET else self.value
-            token = self.argument.tokens[0]
-            provided_by = "" if not token.source or token.source == "cli" else f' provided by "{token.source}"'
-            name = token.keyword if token.keyword else self.argument.name.lstrip("-").upper()
-            message = f'Invalid value "{value}" for "{name}"{provided_by}.'
+            try:
+                token = self.argument.tokens[0]
+            except IndexError:
+                message = f'Invalid value "{value}" for "{self.argument.name}".'
+            else:
+                provided_by = "" if not token.source or token.source == "cli" else f' provided by "{token.source}"'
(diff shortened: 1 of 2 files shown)


wohali:
Thanks for looking at this so quickly.

This is getting me part of the way there. However, the formatting isn't great. From my project code:

```
╭─ Error ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Invalid value "CommonArgs(debug=True, verbose=False, out=PosixPath('t'), workers=4)" for "*". Unable to create debug directory     │
│ /home/wohali/sample/t/debug                                                                                         │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

While the actual raised `ValueError` custom exception text is being shown, I'm not sure leaking the internal class name, its constructor values or the `*` scope to the end user is anything but confusing.

Still, this fixes the problem. It has to work before it has to look pretty. 😉 

My other issue around validators is more stylistic, that is, dealing with `ValueError`s when no common arguments are passed (so the function receives None), runs the constructor, and fails -- in my case, inside of another decorator (that is setting up logging). It's unrelated and I've tied my brain in knots thinking about it, so, I'll come back to that later.

BrianPugh:
I think I should certainly special case the name `"*"` to not show in the error message.

And also I think I agree with you that exposing the coerced value in this case is probably not too meaningful for the end-user. I think in this case it's best to just propagate along the error-message from the raised exception.

BrianPugh:
I guess there's no need to special case `"*"` since i won't be printing the value anyways 🙈

BrianPugh:
alright, please give this a try now! Basically it just removes all the automatic bits and propagates along the ValueError message. I agree with you that all the other bits probably don't mean much to the end user and thusly should not be displayed.

wohali:
```
tests/test_cli.py::test_broken_debug 
╭─ Error ────────────────────────────────────────────────────────────────────────────────╮
│ Unable to create debug directory /home/wohali/samples/t/debug            │
╰────────────────────────────────────────────────────────────────────────────────────────╯
PASSED                                        [100%]
```

👍🏼 +1 thank you!

Issue #308, opened 2025-01-24
Name of @app.default function appears in help


BruceEckel:
In the first example in https://cyclopts.readthedocs.io/en/latest/getting_started.html:
```python
from cyclopts import App

app = App()

@app.default
def main():
    print("Hello World!")

if __name__ == "__main__":
    app()
```

If we use `-h`, we get:
```text
Usage: main COMMAND

╭─ Commands ───────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --help -h  Display this message and exit.                                                                            │
│ --version  Display application version.                                                                              │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```
But exposing the name `main` is confusing in this case, because IMO the point of `@app.default` is to not require the user to use or even know the name of the default function. I could not find any arguments for `default` that would modify this behavior.

I can hack it by using `_` as the function name:
```python
from cyclopts import App

app = App()


@app.default
def _(strings: list[str] = None):
    if strings:
        print(f"{strings = }")
    else:
        print("no args")


app()
```
Now when I use `-h` I get:
```text
Usage:  COMMAND [ARGS] [OPTIONS]

╭─ Commands ──────────────────────────────────────────────────────────────────────────────────╮
│ --help -h  Display this message and exit.                                                   │
│ --version  Display application version.                                                     │
╰─────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Parameters ────────────────────────────────────────────────────────────────────────────────╮
│ STRINGS --strings --empty-strings                                                           │
╰─────────────────────────────────────────────────────────────────────────────────────────────╯
```
1. Here the `Usage:` seems to indicate that `COMMAND` is required. (Is there a way to modify the `Usage:` string?)
2. `[ARGS]` appears to mean the same as what the help is calling `Parameters`. In this case it would be more usefully indicated as `[STRINGS]`
3. There are no `[OPTIONS]` for this program. It's confusing to say that there are, or perhaps `[OPTIONS]` is meant to be help or version?

BrianPugh:
So this is intentional, but perhaps the feature was not super well thought out. When resolving the name of an `App`, the function name of `default` [gets second priority](https://cyclopts.readthedocs.io/en/latest/api.html#cyclopts.App.name). This has some niceties for some subapps, but can definitely seem suboptimal at the root level.

[...]

PR #441, opened 2025-05-19
Only display COMMANDS in the help-page if there are registered commands

BrianPugh:
[The code always intended to do this](https://github.com/BrianPugh/cyclopts/blob/accbea6bf41371bd716a169924cd27ed8d3cabd1/cyclopts/help.py#L342), but broke with the transition to v3 where help/version flags got reclassified as commands.

Addresses bug mentioned [here](https://github.com/BrianPugh/cyclopts/issues/308#issuecomment-2891960743). @taranlu-houzz what do you think of this solution?

## TODO

- [x] fix unit tests.
- [x] Add an explicit unit test for this.

### DIFF
--- a/cyclopts/core.py
+++ b/cyclopts/core.py
@@ -495,6 +495,16 @@ class App:
     def sort_key(self, value):
         self._sort_key = value
 
+    @property
+    def _registered_commands(self) -> dict[str, "App"]:
+        """Commands that are not help or version commands."""
+        out = {}
+        for x in self:
+            if x in self.help_flags or x in self.version_flags:
+                continue
+            out[x] = self[x]
+        return out
+
     def version_print(
         self,
         console: Optional["Console"] = None,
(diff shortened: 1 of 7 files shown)


taranlu-houzz:
@BrianPugh Will try to test soon. Will this handle the situation where you register a single command to the app and make it the default?

BrianPugh:
It should; simple test example:
```python
from cyclopts import App


app = App()


@app.default
def main(*, arg1):
    print(arg1)


if __name__ == "__main__":
    app()
```
```bash
$ python issue-308.py --help
Usage: main [OPTIONS]

╭─ Commands ───────────────────────────────────────────────────────────────────╮
│ --help -h  Display this message and exit.                                    │
│ --version  Display application version.                                      │
╰──────────────────────────────────────────────────────────────────────────────╯
╭─ Parameters ─────────────────────────────────────────────────────────────────╮
│ *  --arg1  [required]                                                        │
╰──────────────────────────────────────────────────────────────────────────────╯
```

Issue #445, opened 2025-05-27
[Feature request] A way to set `None` value


zumm:
Would be nice to have a way to set `None` value. Something like this:

```python
# cli.py
from pathlib import Path
from cyclopts import App

app = App()


@app.default
def main(
    *,
    some_file: Path | None = Path("some_file.txt"),
):
    print(f"{some_file=}")


if __name__ == "__main__":
    app()
```

```bash
> uv run cli.py --some-file none
some_file=None
```

According to https://clig.dev this is important feature.

BrianPugh:
this one is a bit tough; some thoughts in no particular order:
1.  Having special values that are within the "solution space" of a normal user response (e.g. a path named "none") could result in a hard-to-find bug.
2. An explicit sentinel value like `none` smells like an abstraction leakage in this context; `None` is a common default (especially for mutable or optional values), but `None` frequently doesn't have a direct meaning for the CLI end-user.



Simultaneously, I do understand that sometimes people need this. How do you feel about this solution in your use-case:

```python
from pathlib import Path
from typing import Annotated

from cyclopts import App, Parameter

app = App()


def optional_path(type_, tokens) -> Path | None:
    value = tokens[0].value
    return None if value == "none" else type_(value)


OptionalPath = Annotated[Path | None, Parameter(converter=optional_path)]


@app.default
def main(
    *,
    some_file: OptionalPath = Path("some_file.txt"),
):
    print(f"{some_file=}")


if __name__ == "__main__":
    app()
```

```bash
$ python issue-445.py --some-file none
some_file=None

$ python issue-445.py --some-file foo.bin
some_file=PosixPath('foo.bin')

$ python issue-445.py
some_file=PosixPath('some_file.txt')
```

zumm:
I agree that negative flag solution is perfectly aligned with other negative flag features. For my liking negative flags bloat help and harder to use in automatization, but as long as it solves problem i'm ok with it.

BrianPugh:
released in v3.17.0

PR #446, opened 2025-05-28
Add `Parameter.negative_none` feature.

BrianPugh:
Addresses #445.

@zumm can you give this a try? Thanks!

### DIFF
--- a/cyclopts/annotations.py
+++ b/cyclopts/annotations.py
@@ -105,7 +105,9 @@ def is_typeddict(hint) -> bool:
     )
 
 
-def resolve(type_: Any) -> type:
+def resolve(
+    type_: Any,
+) -> type:
     """Perform all simplifying resolutions."""
     if type_ is inspect.Parameter.empty:
         return str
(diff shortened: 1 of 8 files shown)


zumm:
I have tested it and it works. Thx!

Althrough after second thougth and some ux testing, negative flags don't look good for some of my cases. For example, there is required option "email" in my app. If it isn't set app will prompt it. In this case flag "--no-email"/"--none-email" is confusing. Users misinterpret it as "app will work without email" when real meaning is "app will ask email even if email is already set in config file".

BrianPugh:
This was my initial hesitation of adding a `negative_none` in the first place (but now that it is here, there's probably no harm in merging it). Usually a `None` value is special enough that the meaning is a little hard to describe to the CLI user.

I suppose in your case, you could call it `--prompt-email`?

zumm:
Yep, i think i will add "propmt-" prefix for required (prompted) options and "no-" for not required ones.

BrianPugh:
sounds good! Since we're happy with this solution, I'll find some time early this week to fix whatever reasons CI isn't passing this and cut a release!

Issue #447, opened 2025-05-29
KeyboardInterrupt should not print a traceback


gremlation:
When you press Ctrl-C while a synchronous command is being run, it quits the process and prints a `KeyboardInterrupt` traceback.

When you do it while an asynchronous command is being run, it quits the process and prints two tracebacks, an `asyncio.exceptions.CancelledError` caused by a `KeyboardInterrupt`.

This is very ugly and the user doesn't need to be given a big traceback to diagnose what happened. They know they hit Ctrl-C. The process should just be terminated without printing anything.

PR #448, opened 2025-05-29
Stop `KeyboardInterrupt` from printing tracebacks

gremlation:
Fixes: #447

### DIFF
--- a/cyclopts/core.py
+++ b/cyclopts/core.py
@@ -1251,7 +1251,7 @@ class App:
                 return command(*bound.args, **bound.kwargs)
         except KeyboardInterrupt:
             if self.suppress_keyboard_interrupt:
-                sys.exit(130) # Use the same exit code as Python's default KeyboardInterrupt handling.
+                sys.exit(130)  # Use the same exit code as Python's default KeyboardInterrupt handling.
             else:
                 raise
 


gremlation:
This isn't especially tied to sync vs async, it just affects async worse because the traceback is much larger.

The issue here is that normal operation when cancelling a command-line tool is to exit with a `130` status, not to print lots of debug output in relation to an error. In practical terms, I want to exit out of a running process and read the output of the command so far. Instead that output has been scrolled off the screen by ~100 lines of traceback informing me about the exception that I triggered.

`KeyboardInterrupt` is unique among the exceptions, particularly for command-line tools, because it signals user intent, not an error that needs to be diagnosed. `130` is the normal Ctrl-C exit code, which [Typer uses as well](https://github.com/fastapi/typer/blob/9138340f90258c9ef42a5159fc6bebc7373dbdf8/typer/core.py#L210). I've added a comment.

With the rich tracebacks example, it's not clear to me how this would work when I'm exposing a command-line tool via `pyproject.toml` like this:

```toml
[project.scripts]
foo = "foo:app"
```

I'd have to wrap `app`, but pass through all of its attribute accesses - is this just a case of subclassing then overriding `__call__()`?

BrianPugh:
alright, so I played around with this a little bit and I think I like it. I can make a few tweaks to this PR, but want some of your input first:
1. I'll fix whatever CI is complaining about.
2. I want to add an `App` parameter to configure if this is enabled or not. What do you think of the name `suppress_keyboard_interrupt`? I'll make this `True` by default, because I agree with you that the `KeyboardInterrupt` trace is not typically helpful to be exposed to the end user of a CLI.

I want to do a release by Friday, and would love to have this in for that!
---

To answer/respond to some of your other comments (which will be less applicable after merging this, but still good discussion!):

> With the rich tracebacks example, it's not clear to me how this would work when I'm exposing a command-line tool via pyproject.toml

I typically do something like:

```python
app = App()

def run_app():
    app()
```

```toml
[project.scripts]
foo = "foo:run_app"
```

This is a little bit more verbose, but it makes it easier to have even greater control around app launching and exceptions. Similar things can be accomplished via using a [meta app](https://cyclopts.readthedocs.io/en/latest/meta_app.html), but they're typically a bit clunky/complicated for some situations.

> I'd have to wrap app, but pass through all of its attribute accesses - is this just a case of subclassing then overriding __call__()?

Subclassing and overriding `__call__` would be another option over the `run_app` solution above.

gremlation:
Great - I've added that option. Thanks for the tips on overriding!

BrianPugh:
Thanks! I'll release this within the next few days!

Issue #452, opened 2025-06-02
Combining multiple short flags


seowalex:
Hi, thank you for your work on Cyclopts! I am migrating from Typer and have been very satisfied so far.

One thing which I miss from Typer is the ability to combine multiple short flags (for e.g., `ls -lah`). Could this be added?

BrianPugh:
I think this can be added without too much work; i'll investigate!

BrianPugh:
addressed in v3.17.0!

PR #453, opened 2025-06-03
Parse combined short flags

BrianPugh:
Parses combined short (single character flags) popular in CLI applications. For example, `ls -alh`.

Addresses #452. @seowalex please give this a try!

## TODO

- [x] Add unit test including unknown flags.
- [x] Add unit test to trigger
   ```
   CycloptsError(msg=f"Cannot combine flags and short-options in token {cli_option}")
   ```
- [x] Make a specific exception for ^

### DIFF
--- a/cyclopts/__init__.py
+++ b/cyclopts/__init__.py
@@ -8,6 +8,7 @@ __all__ = [
     "ArgumentOrderError",
     "Token",
     "CoercionError",
+    "CombinedShortOptionError",
     "CommandCollisionError",
     "CycloptsError",
     "CycloptsPanel",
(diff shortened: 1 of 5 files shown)


seowalex:
Thanks for the quick turnaround! Will try and report back.

seowalex:
It seems to work well, thanks! One unrelated problem I found is that single digit flags (like `-1`) do not seem to work. It seems that they are being parsed as positional variables.

BrianPugh:
I'll wrap up this PR for a release, and we can discuss single-digit flags separately. Actually can you start a github issue for that topic? Try to source how other libraries/common-CLIs do it. A primary issue with single digit flags is that it means that the CLI could no longer parse positional integer arguments. But please open a thread for further discussion!

Issue #454, opened 2025-06-05
Short flags with digits do not work


seowalex:
It seems to work well, thanks! One unrelated problem I found is that single digit flags (like `-1`) do not seem to work. It seems that they are being parsed as positional variables.

_Originally posted by @seowalex in https://github.com/BrianPugh/cyclopts/issues/453#issuecomment-2939012254_

seowalex:
After a bit more experimentation, it seems that short flags with any number of digits do not work. One on hand, I think this is the correct default behaviour, since the alternative would prevent negative numbers from being used as positional parameters. On the other hand, I am attempting to mimic `ls`'s interface, which uses `-1` to list one file per line.

Probably we can add some option for this? Although I understand if you do not want to clutter the options with this edge case.

BrianPugh:
Added in v3.18.0!

PR #455, opened 2025-06-06
Allow digit flags.

BrianPugh:
Addresses #454 .

@seowalex I didn't add any options to enable/disable this, because it only applies if someone explicitly gives a single-digit-name to a flag. And at that point, then the developer should inherently understand the consequences if they were also trying to parse a negative integer from the CLI. Could you give this a try? Thanks!


seowalex:
Yes, this works well for me. I must add that your turnaround time for these small feature requests have been fantastic, thank you very much!

PR #472, opened 2025-06-22
config.Env: recombine remaining_keys into a single key if the found argument doesn't have any sub-arguments.

### LINKED ISSUE #471: [BUG] `cyclopts.config.Env` is not parsed correctly in the presence of `**kwargs`
(Feel free to edit the issue title, because I am not sure that I found the root cause correctly)

Consider the following `script.py`:

```
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "cyclopts==3.19.0",
# ]
# ///

import cyclopts


def foo_impl(a: str, b: str):
    print(locals())


def foo(a: str, **kwargs):
    return foo_impl(a, **kwargs)


app = cyclopts.App(config=[cyclopts.config.Env('BAR_', command=False)])
app.default(foo)
app()

```

This runs as expected and results in the exception telling that `b` is missing:

```
BAR_ONEWORD=123 uv run local/a.py --a A
```

However, the following fails during the argument parsing because of the switch from `BAR_ONEWORD` to `BAR_TWO_WORDS`

```
BAR_TWO_WORDS=123 uv run local/a.py --a A
```

Error:

```
╭─ Error ─────────────────────────────────────────────────────────────────────────────────────────────╮
│ Dictionary of tokens provided for unknown <class 'str'>.                                            │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

BrianPugh:
Should resolve #471. @yura52 can you give this a try?

### DIFF
--- a/cyclopts/config/_env.py
+++ b/cyclopts/config/_env.py
@@ -40,18 +40,27 @@ class Env:
 
         candidate_env_keys = [x for x in os.environ if x.startswith(prefix)]
         candidate_env_keys.sort()
+        delimiter = "_"
         for candidate_env_key in candidate_env_keys:
             try:
                 argument, remaining_keys, _ = arguments.match(
                     candidate_env_key[len(prefix) :],
                     transform=_transform,
-                    delimiter="_",
+                    delimiter=delimiter,
                 )
             except ValueError:
                 continue
             if set(argument.tokens) - added_tokens:
                 # Skip if there are any tokens from another source.
                 continue
+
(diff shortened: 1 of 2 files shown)


Yura52:
Hi @BrianPugh , thanks for the quick fix! I have just tested the patch locally and it fixes the minimal example that I provided in the issue.

BrianPugh:
This will go into the next release, hopefully later today!

Issue #470, opened 2025-06-18
Group validator `ArgumentCollection` content inconsistency and call inconsistency.


BrianPugh:
I think my original intention with Group Validators was that:
1. The `ArgumentCollection` provided to the validator contains all the arguments in the group (whether they were set or not).
2. The Group validator is called no matter what.

Currently I think both of these points are implemented incorrectly:
1. Only arguments that were CLI-specified are in the provided Argument Collection.
2. The group validator is only called if at least 1 argument within the group is provided.

BrianPugh:
fixed by #473

PR #473, opened 2025-06-22
Always call all group-validators; provide all arguments within the group.

BrianPugh:
Addresses #470. Tagging @hans-d as the original reporter; it would be appreciated if you could give this PR a try and/or provide feedback!

### DIFF
--- a/cyclopts/bind.py
+++ b/cyclopts/bind.py
@@ -328,7 +328,7 @@ def _sort_group(argument_collection) -> list[tuple["Group", ArgumentCollection]]
     out = {}
     # Sort alphabetically by group-name to enfroce some determinism.
     for i, group in enumerate(sorted(argument_collection.groups, key=lambda x: x.name)):
-        if not (group_arguments := argument_collection.filter_by(group=group, has_tree_tokens=True)):
+        if not (group_arguments := argument_collection.filter_by(group=group)):
             continue
         common_root_keys = _common_root_keys(group_arguments)
         # Add i to key so that we don't get collisions.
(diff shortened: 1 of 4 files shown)


hans-d:
Thanks. This does change some behaviour, but works for the better now. Thanks!

hans-d:
hmmm... as we get all the fields, it is required to do some filtering for the group specific fields... which cannot refer to the group var as the group var depends on the validator...

hans-d:
perhaps the first arg should just contain the fields of the group the validator is attached to, and an optional second arg for the all the arguments?

BrianPugh:
>mmm... as we get all the fields, it is required to do some filtering for the group specific fields...

The provided `ArgumentCollection` should only contain the arguments that belong to the group; are you seeing something else?

If you have a simplified script you could post, it would be good to share so we can see if things are behaving properly!

hans-d:
let me double check, I was using some workarounds that may have caused the problem.

hans-d:
ok, works as expected. My mistake.
As now all the parameters are provided, I need to do an explicit value_set check.
The filter by seems to go wrong, as `pprint(args.filter_by(value_set=True))` and `pprint(args.filter_by(value_set=False))` return the same result showing ` _value=<UNSET>` with the pprint. (not seeing a test for this as well). WIll investigate further.

hans-d:
(or should I use has_tokens, that seems to be working)

BrianPugh:
linking https://github.com/BrianPugh/cyclopts/issues/479#issuecomment-3006795709 because some relevant discussion is happening over there.

Issue #467, opened 2025-06-17
Improve help output


daviewales:
I have a command signature such as the following:

``` python
import cyclopts

def main(one: str, /, *, two: str | None = None):
    """Example command

    Args:
        one: An argument

    Params:
        two: An optional parameter
    """
    print(f"{one=}")
    print(f"{two=}")

if __name__ == '__main__':
    cyclopts.run(main)
```

## Without arguments

When run without arguments, I get the following:

```
python temp.py
╭─ Error ──────────────────────────────────────────────────╮
│ Parameter "ONE" requires an argument.                    │
╰──────────────────────────────────────────────────────────╯
```

This could be improved with a hint that I should use the `-h` or `--help` flags to get help.

## With `-h`

When run with `-h`, I get the following:

```
python temp.py -h
Usage: main [ARGS] [OPTIONS]

Example command

╭─ Arguments ──────────────────────────────────────────────╮
│ *  ONE  An argument [required]                           │
╰──────────────────────────────────────────────────────────╯
╭─ Commands ───────────────────────────────────────────────╮
│ --help -h  Display this message and exit.                │
│ --version  Display application version.                  │
╰──────────────────────────────────────────────────────────╯
╭─ Parameters ─────────────────────────────────────────────╮
│ --two  An optional parameter                             │
╰──────────────────────────────────────────────────────────╯
```

Firstly, I _love_ that this automatically picks up my documentation from the docstring, without needing to contort my function signature.
However, I feel that the _Arguments_ and _Parameters_ boxes should be neighbours. I don't like that the `Commands` box appears between them.

Also, I feel that at least from the perspective of the end user, `--help` and `--version` are parameters (optional arguments?), rather than commands?

**Edit**:

One more thing. If I add a section to my docstring called `Examples:`, it disappears from the help! If I remove the colon, it appears as expected.

BrianPugh:
> This could be improved with a hint that I should use the -h or --help flags to get help.

We could tack on a hint like "See --help for usage information" to each error message (maybe just some error messages?). But some people might consider it unnecessary clutter. Thoughts?

---

> However, I feel that the Arguments and Parameters boxes should be neighbors. I don't like that the Commands box appears between them.

[...]

PR #476, opened 2025-06-24
Give Commands/Arguments/Parameters higher priority (in that order) on the help-page.

BrianPugh:
Now the ordering of the help-page will always go:

```
Commands
Arguments
Parameters
Other Custom User Groups (sorted and unsorted)
```

This partially addresses #467. I think in general this results in better ordering.  @daviewales can you play around with this PR and see if you better like the help-age?


daviewales:
Thanks! Yes, I like this better. I'm still not sure how I feel about `--help` being a command. But I kind of get it.
Argparse has `--help` as an option:
https://docs.python.org/3/howto/argparse.html#argparse-tutorial

PR #487, opened 2025-07-02
Add check for `Annotated` to `_get_choices` in help generation

Rubikoid:
Initial problem was generating help for annotated enums with typealias. 

Maybe better to fix this problem in function, that unwraps `Annotated` when it passed directly to function params, without typealias?


BrianPugh:
fix looks good to me! if we run into additional, similar problems, we can investigate solving it at other locations in the code-base. Thank you!

BrianPugh:
released in v3.22.1

Issue #491, opened 2025-07-07
[BUG] Empty lists in configs are treated as missing arguments


Yura52:
Hi, here is a minimal example:

```python
# app.py
import cyclopts

def f(values: list[int]) -> None:
    print(values)

app = cyclopts.App(config=cyclopts.config.Toml('config.toml'))
app.default(f)
app()
```

```toml
# config.toml
values = []
```

```
$ python app.py
...
Parameter "--values" requires an argument.
```

BrianPugh:
thanks for the bug report and easy-to-reproduce example!

BrianPugh:
Fixed in v3.22.2

PR #492, opened 2025-07-07
fix erroneous MissingArgumentError if an empty iterable value is provided in a config file.

BrianPugh:
Addresses #491. @Yura52 can you give this a try? Thanks!


Yura52:
Hi! I don't know conventient ways to quickly test suggested changes, and inconventient ways take some time, so I don't know when I will test this. If there is a set of commands that I can copy & paste in my terminal, I am ready to do that and share the output.

BrianPugh:
i'll just merge it in :D 

To test it in your project, you can just clone this repo and run `pip install -e .` and it will install it to your environment.

Issue #494, opened 2025-07-23
`--version` in entry point


whatever60:
Thank you for creating this amazing package!

I am having a question regarding the automatic imputation of package version based on reflection, when the app is launched in entry points specified in `pyproject.toml`.

Let's say we use `uv init`, and create a `cli.py`, and a function `main`, and we put in `pyproject.toml` file

```toml
[project.scripts]
"my_package" = "my_package.cli:main"
```

The `_get_root_module_name` will traverse the following modules:

```
<module 'cyclopts.core' from '.venv/lib/python3.10/site-packages/cyclopts/core.py'>
<module 'cyclopts.core' from '.venv/lib/python3.10/site-packages/cyclopts/core.py'>
<module 'cyclopts.core' from '.venv/lib/python3.10/site-packages/cyclopts/core.py'>
<module 'cyclopts.core' from '.venv/lib/python3.10/site-packages/cyclopts/core.py'>
<module '__main__' from '.venv/bin/my_package'>
```

and since the last module is automatically created by the build system based on entry points under in the toml file, it won't have a `__version__`.

Right now (to the best of my knowledge), we'd have to explictly set `version` in every single `cyclopts.App` exposed as entry point. But since `.venv/bin/my_package` does have `from my_package.cli import main`, maybe we can turn  `_default_version` from a function to a method of the `App` class. It first detects version based on the currect logic, then on the root import module of the current app instance, and finally the default 0.0.0.

Any tips or feedback would be awesome!

BrianPugh:
thanks for opening this issue and the PR! We'll continue discussion in #495

whatever60:
Solved in #495.

Cheers

BrianPugh:
Fixed in v3.22.3

PR #495, opened 2025-07-23
Include root import module in automatic version detection

whatever60:
--version now automatically check the version of root import module (e.g. `my_package` of from `my_package.cli import main`) before returning default value

The logic of importing from importlib is now in a global manner instead of inside the function _default_version. In App.version_print, when version is None or 0.0.0, we now tries to check the version of the root module that instantiated the App object. That is, if app is instantiated in the module my_package.cli, App.version_print will try to get the __version__ the package my_package.

This change is mainly to enable automatic version detection in entry points. tests are performed with pytest. resulting in 19 failed, 837 passed, 4 skipped both before and after this commit.

### DIFF
--- a/cyclopts/core.py
+++ b/cyclopts/core.py
@@ -1,3 +1,4 @@
+import importlib
 import inspect
 import os
 import sys
@@ -9,6 +10,7 @@ from enum import Enum
 from functools import lru_cache, partial
 from itertools import chain
 from pathlib import Path
+from types import ModuleType
 from typing import (
     TYPE_CHECKING,
     Annotated,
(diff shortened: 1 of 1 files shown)


whatever60 (commented):
#494

BrianPugh:
thanks for opening the issue, the kind words, and for the follow up PR! I need to look at this a little closer; i think that this might introduce nontrivial slow-downs. For example, on my machine running all tests goes from 2.37s to 8.46s.

whatever60:
Great catch! I am seeing this too on my machine. I changed inspect.stack to sys._getframe for getting the specific frame and the performance difference of pytest is negligible.

cyclopts/core.py
```diff
@@ -531,8 +546,15 @@ def version_print(
 
         version_raw = self.version() if callable(self.version) else self.version
 
-        if version_raw is None:
+        if version_raw is None or version_raw == "0.0.0":
```
BrianPugh: do we need this check? If the user explicitly specifies `App(version="0.0.0")`, we should respect that version string.

BrianPugh: ah, it's because how the current code is structured; let me think for a moment.

whatever60: yeah maybe we can make `_default_version` return None instead of 0.0.0 by default, as the `App.version_print` method also checks for None value. Not sure whether any test cases need to be changed though if we go that route.

BrianPugh: I just made a commit changing this; let me know what you think!

whatever60:
Thank you for the fix. I think that works, with the caveat that the new logic now precedes the previous logic.

BrianPugh:
I think that's fine; i'll just update the documentation and then merge this!

BrianPugh:
so actually, the behavior you implemented was always the intended behavior and pretty much is what is documented, so I don't think any documentation actually needs to change. Thanks for this PR!

Issue #508, opened 2025-07-31
Command `alias` is unknown if  `name` is provided


ottowayi:
If you provide both the `name` and `alias` params to the the `command` decorator, then the `alias` is not recognized.

poc:

```py
# /// script
# dependencies = [
#   "cyclopts",
# ]
#  requires-python = ">=3.12"
# ///

from cyclopts import App

app = App()


@app.command(name='list', alias='ls')
def list_something():
    print('list')

@app.command(alias='show')
def display():
    print('display')


if __name__ == '__main__':
    app()
```

When run, the alias `show` correctly works for `display`, but the alias `ls` for `list` is not recognized:

```pwsh
~/code/playground> uv run ./cyclopts-cmd-alias-poc.py list
Installed 9 packages in 2.03s
list
~/code/playground> uv run ./cyclopts-cmd-alias-poc.py ls  
╭─ Error ──────────────────────────────────────────────────────────────────────────────╮
│ Unknown command "ls". Did you mean "list"? Available commands: list, display, show.  │
╰──────────────────────────────────────────────────────────────────────────────────────╯
~/code/playground> uv run ./cyclopts-cmd-alias-poc.py display
display
~/code/playground> uv run ./cyclopts-cmd-alias-poc.py show   
display
```

BrianPugh:
thanks for the reproducible, terse example! I opened up a fix in #509.

BrianPugh:
Fixed in v3.22.5!

PR #509, opened 2025-07-31
fix broken command alias when name is also specified.

BrianPugh:
Addresses #508 

@ottowayi can you please give this a try? thanks!


ottowayi:
yep, worked for my poc using `uv run --with "cyclopts @ git+https://github.com/BrianPugh/cyclopts@bugfix/command-name-and-alias"` 

thanks for the fast response on this

Issue #507, opened 2025-07-31
Cannot parse list of dataclass


VRichardJP:
Hi!

It looks like it is not possible to use a list of dataclasses as argument (e.g. `list[User]`). Is it supposed to be supported? For example using a YAML config file:

```python
from pathlib import Path
from dataclasses import dataclass
from typing import Annotated, Literal

import cyclopts.config
from cyclopts import App, Parameter

app = App()


@dataclass
class User:
    name: str
    age: int
    region: Literal["us", "ca"] = "us"


@app.default
def main(users: list[User]):
    print(locals())


@app.meta.default
def meta(
    *tokens: Annotated[str, Parameter(show=False, allow_leading_hyphen=True)],
    config: Annotated[Path | None, Parameter(help="Application configuration as YAML file")] = None,
):
    # Can load application configuration from yaml file provided by --config
    if config is not None:
        app.config = cyclopts.config.Yaml(config)  # type: ignore
    app(tokens)


if __name__ == "__main__":
    app.meta()

```

Running with a single item in the YAML config gives an empty list:
```
$ cat ./test.yaml
users:
  - name: alice
    age: 22
    region: us
$ python ./test.py --config ./test.yaml
{'users': []}
```

Running with multiple items fails with an error:
```
$ cat ./test.yaml
users:
  - name: alice
    age: 22
    region: us
  - name: bob
    age: 33
    region: ca
$ python ./test.py --config ./test.yaml
╭─ Error ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Invalid value for "[users]" from /home/vrichard/zenoh_everfleet_bridge/test.yaml: unable to convert "{'name': 'bob', 'age':    │
│ 33, 'region': 'ca'}" into int.                                                                                                │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

BrianPugh:
Hi @VRichardJP!

Currently Cyclopts does not support parsing of iterable dataclass-like structures. Namely, assigning keyword-specified values to a list of items. We do support iterable tuples of definite length (e.g. `list[tuple[str, int]]`). This is due to the strong ambiguity of specifying parameters from the CLI.

Ideally we should make an exception that makes this more obvious to the developer.

Welcoming any thoughts or feedback!

VRichardJP:
Thanks for the quick feedback! I have not peeked at the implementation, but an exception would definitely help :)

[...]

PR #510, opened 2025-08-01
raise exception if a list[dict] is attempted to be used from a config file.

BrianPugh:
Addresses #507.


BrianPugh:
rendered obsolete in v4-develop

PR #517, opened 2025-08-11
docs: improve `consume_multiple`

nachocab:
This library is such a joy to use. Thank you so much for creating it!

I was reading through the docs and found a few things that could be improved. Let me know if you'd like me to change something:

- improve `Parameter.consume_multiple` and "Keyword Multiple Values"
  - Fix paragraph (“or an and”), unify with the corresponding language in `rules.rst`, add a positive example and simplify the negative example. Add `:obj:` to `False`
  -  unify the examples in `getting_started.rst` and `vs_typer/README.rst`

- add syntax to code blocks
- Improve link in `interactive_help.rst` (it shouldn’t be monospaced)
- Simplify paragraph in `getting_started.rst`

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -84,7 +84,7 @@ API
          app = App(help="This is my help string.")
          app()
 
-      .. code-block::
+      .. code-block:: console
 
          $ my-script --help
          Usage: scratch.py COMMAND
@@ -1151,7 +1151,7 @@ All definitions in this section are simply predefined annotations for convenienc
 
 Due to Cyclopts's advanced :class:`.Parameter` resolution engine, these annotations can themselves be annotated to further configure behavior. E.g:
 
-.. code-block::
+.. code-block:: python
 
    Annotated[PositiveInt, Parameter(...)]
 
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -749,7 +749,7 @@ API
       :type: Optional[bool]
       :value: None
 
-      If ``False``, treat the user-defined class annotation similar to a tuple.
+      If :obj:`False`, treat the user-defined class annotation similar to a tuple.
       Individual class sub-parameters will not be addressable by CLI keywords.
       The class will consume enough tokens to populate all required positional parameters.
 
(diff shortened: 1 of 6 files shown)


docs/source/api.rst
```diff
@@ -828,9 +828,27 @@ API
       :type: Optional[bool]
       :value: None
 
-      When a parameter is **specified by keyword**, consume multiple elements worth of CLI tokens.
-      Will consume tokens until the stream is exhausted, or an and :attr:`.allow_leading_hyphen` is False
-      If ``False`` (default behavior), then only a single element worth of CLI tokens will be consumed.
+      If :obj:`True`, consume multiple elements worth of CLI tokens until the stream is exhausted (or, if :attr:`Parameter.allow_leading_hyphen` is :obj:`False`, until an option-like token is reached).
```
BrianPugh: I think we still need to mention that this only applies when the parameter is specified by `--keyword`.

nachocab: I've reworked the example for clarity, but in some cases `consume_multiple` works even if `ext` is not specified by keyword:

```python
@app.command
def name_ext(
    name: str,
    ext: Annotated[list[str], Parameter(consume_multiple=True)],
):
    for extension in ext:
        print(f"{name}.{extension}")

@app.command
def ext_name(
    ext: Annotated[list[str], Parameter(consume_multiple=True)],
    name: str,
):
    for extension in ext:
        print(f"{name}.{extension}")
```

The outputs are:
```shell
$ my-app name-ext "my_file" "txt" "pdf" 
my_file.txt
my_file.pdf

$ my-app ext-name "txt" "pdf" --name "my_file"
my_file.txt
my_file.pdf
```

When it doesn't work, the message is pretty clear:

```shell
$ my-app ext-name my_file --ext "txt" "pdf" 
╭─ Error ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Cannot specify token 'my_file' positionally for parameter 'name' due to previously specified keyword '--ext'. '--ext' must either be passed positionally, or 'my_file' must be  │
│ passed as a keyword to '--name'.                                                                                                                                                │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

```

nachocab: Also, if `consume_multiple` is `False`:

```python
@app.command
def name_ext_single(
    name: str,
    ext: list[str],
):
    for extension in ext:
        print(f"{name}.{extension}")
```

Then, it's surprising that the first call works but not the second or third (I thought they should all fail):

```shell
$ my-app name-ext-single "my_file" "abc" "def"
my_file.abc
my_file.def

$ my-app name-ext-single --name "my_file" "abc" "def"
╭─ Error ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Cannot specify token 'abc' positionally for parameter 'ext' due to previously specified keyword '--name'. '--name' must either be passed positionally, or 'abc' must be passed  │
│ as a keyword to '--ext'.                                                                                                                                                        │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

$ my-app tls name-ext-single --name "my_file" --ext "abc" "def"
╭─ Error ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Unused Tokens: ['def'].                                                                                                                                                         │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

```

nachocab: Ok, never mind. I hadn't realized `consume_multiple` is ignored when the parameter is specified positionally. See my latest commit for the updated explanation/example.

BrianPugh:
This looks good! Thanks for the update! I'll merge it soon.

Issue #346, opened 2025-03-07
Disabling boxes/tables in help output


jscholes:
I was recently made aware of Cyclopts in this [Mastodon post](https://mastodon.social/@tintvrtkovic/114105685003053706), and really like the look of it!  However, I'm running into a bit of a blocker for my use case.

I myself am a blind [screen reader](https://en.wikipedia.org/wiki/Screen_reader) user, and it's important that the apps I develop are as inclusive/accessible as possible.  For people who can see, this means that the default help output with boxes, rounded corners, tables and the like can be legitimately helpful.

But with screen reading software, the characters don't really translate at all into understandable speech or braille output, and it makes the usage, help and error information generated by Cyclopts quite challenging to read.

What I'd really like is to be able to configure Cyclopts to use completely plain text output, or better yet my own custom help formatter, perhaps in response to a particular environment variable, CLI parameter or config file flag being set.  For example with this program:

```
import cyclopts


app = cyclopts.App()


@app.command
def init():
	"""Initialise a thing."""

	print('Initialised!')


@app.command
def export():
	"""Export a thing."""

	print('Exporting...')


if __name__ == '__main__':
	app()
```

The default help output looks something like this:

```
Usage: main.py COMMAND

┌─ Commands ──────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ export     Export a thing.                                                                                          │
│ init       Initialise a thing.                                                                                      │
│ --help -h  Display this message and exit.                                                                           │
│ --version  Display application version.                                                                             │
└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

I'd love to make it look something like this under the appropriate circumstances:

```
Usage: main.py COMMAND

Commands:
export: Export a thing.
init: Initialise a thing.
--help or -h: Display this message and exit.
--version: Display application version.
```

This probably looks quite ugly/basic, but it is pretty screen-reader-friendly.  This is also a very simple program and there'd obviously be a lot of additional formatting to handle in real world scenarios.

Let me know if I can provide more information.  Thanks for sharing your work with us all!

BrianPugh:
Hi @jscholes!

Unfortunately, currently there isn't really a way to customize the layout of the help-page. This is something that we can add to the future; a first step would be to abstract out the data-gathering from the layout-drawing, which would also probably help out #180.

[...]

Issue #180, opened 2024-06-10
[feature request]: automatic markdown documentation generation


danielgafni:
A pretty useful feature Typer [has](https://typer.tiangolo.com/tutorial/typer-command/#generate-docs) is automatic documentation generation. It's really helpful to have this as part of `pre-commit` hooks and in the CI pipeline. 

I wonder if `cyclopts` could also have this?

BrianPugh:
I think we can do that! I think a limitation is that the output docs format will have to be the same format as `App.help_format`. I'll try and get to this!

doodla:
@BrianPugh This feature would be very useful to us as we're building out a new CLI app. Were you able to chip away at this?

BrianPugh:
I haven't started this effort yet; I'll begin it after I finish up my dictionary-support efforts.

danielgafni:
Hey @BrianPugh, do you need help with this? 
I would really like to switch to Cyclopts in my projects, this is the only thing holding me back. 
Let me know if I could do anything.

BrianPugh:
Thanks for offering! I'm in the middle of a big refactor that should make this feature pretty easy for a v3.0.0 cyclopts. Still targeting about a month away.

adam-moss:
I've been looking at for the capability too to switch away from typer entirely.  Current we use https://github.com/bruce-szalwinski/mkdocs-typer for the doc's gen rather than typer's own, I would if a similar plugin would be possible.

BrianPugh:
this hasn't fallen of my radar; I'm just currently prioritizing getting CLI completions out, then I can work on this!

danielgafni:
Hey @BrianPugh, how is it going? Offering my help again.

ESPR3SS0:
Hello all, 

This is a feature that would help me out a ton too; I've made a crude doc generator that I use for my other projects. 

The Fork is here: https://github.com/ESPR3SS0/cyclopts

I say crude because it relies on using `inspect` to pull from the function signatures, and is fairly limited compared to the output from `--help`. I image a true implementation would lean on some of the parsing logic already in the repo... 

With that in mind, I'd enjoy hacking away at a true implementation @BrianPugh . Let me know if I should ditch `inspect`, or if there's any direction you think the implementation should go. 

For example the program: 
```python
from cyclopts import App
from cyclopts.doc_gen import init_docs, generate_docs
from enum import Enum

app = App()

class MyType(Enum):
    OPT1 = "opt1"
[...]

Issue #50, opened 2024-01-03
Support in Trogon


mgielda:
Recently I discovered Trogon (https://github.com/Textualize/trogon) which seems pretty incredible (look at their video demo, this is some great stuff). In short, it turns a well-defined CLI app into a discoverable TUI which shows you all the options, defaults, types etc, lets you choose from available choices etc.

Especially useful for large apps, less so for trivial ones. Still, I think it kind of shows why standardizing your argument parsing etc makes sense, and I am fully on board with cyclopts' ambition to do that using plain Python.

Trogon wraps around click, yet there is some success people have reported with Typer, and an ongoing [effort towards enabling mainline support + docs](https://github.com/Textualize/trogon/pull/14). Trogon+Typer does not fully work for me yet ([it crashes unexpectedly when trying to execute the resulting command](https://github.com/Textualize/trogon/issues/70)), and I suspect this has to do with some minor differences between vanilla Click and Typer. In my disenchantment with Typer recently and having decided I was switching over to Cyclopts, I will most likely not get involved there.

But Trogon remains a cool tool and there [are efforts to make it work with other stuff including e.g. argparse](https://github.com/Textualize/trogon/pull/38) - BTW this is from the author of https://www.fresh2.dev/r/yapx/ which builds on top of Argparse and seems like a good tool to make a comparison against. That effort in Trogon has stalled a bit, and the yapx developer seems to have [forked Trogon](https://github.com/fresh2dev/argparse-tui) to provide TUI support for his argparse-based lib, but I'd potentially be happy to help resurrect the effort to make Trogon more generic and support adding Cyclopts.

As far as comparisons are concerned, besides yapx, it might also make sense to look at [clize](https://github.com/epsy/clize), but I digress...

I guess the main purpose of the issue is to ask you, do you think having support for Cyclopts in Trogon would be useful? I wouldn't want to push for it if you a) found this to be useless and/or b) thought it was fundamentally impossible to do due to the way Cyclopts is implemented as compared to Click/argparse (which I doubt). It most likely won't really require any changes to Cyclopts itself.

And a bit of nostalgia, the first time when I realized how awesome building CLIs in Python can be some years ago was while reading the [plac docs for advanced usage](https://plac.readthedocs.io/en/latest/#advanced-usages-of-plac) and all the nifty tricks described there. I was going to ask what your thoughts were on adding an interactive prompt (which I think is the niftiest of the nifty tricks)... until I found you'd already added it. Awesome!

[...]

Issue #497, opened 2025-07-23
`default_parameter` ignored when defining a meta app.


BrianPugh:
> Relatedly, while trying to debug this I found that defining a meta app causes the `default_parameter` to be ignored in sub-apps: In my case, I globally disable negative flags with `default_parameter=Parameter(negative=[])`, and after adding a meta app, the sub-apps no longer respect this. 

 _Originally posted by @seowalex in [#496](https://github.com/BrianPugh/cyclopts/issues/496#issuecomment-3106865047)_

seowalex:
# Without meta App

```python
from cyclopts import App, Parameter

app = App(default_parameter=Parameter(negative=[]))


@app.command
def foo(*, my_other_flag: bool = False):
    print(my_other_flag)


@app.default
def main(*, my_flag: bool = False):
    print(my_flag)


if __name__ == "__main__":
    app()
```

```python
$ uv run test.py -h
Usage: main COMMAND [OPTIONS]

╭─ Commands ─────────────────────────────────────────────────────────────────────────────────────╮
│ foo                                                                                            │
│ --help -h  Display this message and exit.                                                      │
│ --version  Display application version.                                                        │
╰────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Parameters ───────────────────────────────────────────────────────────────────────────────────╮
│ --my-flag  [default: False]                                                                    │
╰────────────────────────────────────────────────────────────────────────────────────────────────╯
```

```python
$ uv run test.py foo -h
Usage: main foo [OPTIONS]

╭─ Parameters ───────────────────────────────────────────────────────────────────────────────────╮
│ --my-other-flag  [default: False]                                                              │
╰────────────────────────────────────────────────────────────────────────────────────────────────╯
```

# With meta App

```python
from typing import Annotated

from cyclopts import App, Parameter

app = App(default_parameter=Parameter(negative=[]))


@app.command
def foo(*, my_other_flag: bool = False):
    print(my_other_flag)


@app.default
def main(*, my_flag: bool = False):
    print(my_flag)


@app.meta.default
def launcher(
    *tokens: Annotated[str, Parameter(show=False, allow_leading_hyphen=True)],
    verbose: Annotated[list[bool], Parameter(alias="-v")] = [],
):
    verbosity = sum(verbose)
    print(f"{verbosity=}")

    app(tokens)


if __name__ == "__main__":
    app.meta()
```

```python
$ uv run test.py -h
[...]

PR #518, opened 2025-08-14
Abstract help

ESPR3SS0:
# Abstraction of HelpPanel Generation
Divided the logic for the HelpPanel Generation, seperating how the panel is formatted / rendered from data gathering. 

## TL;DR

1. Added Spec Objects: `ColumnSpec`, `TableSpec`, `PanelSpec` to control how the Panel Looks 
2. Added Entry Objects: `AbstractTableEntry`, `TableData`, to all custom members to be included in an entry 
3. Should / may be helpful with #180 #50 #346 

## General Workflow for HelpPanel Generation 

* Define a list of `ColumnSpecs`; Each spec is responsible for defining the look of the column, as well as rendering each entry value in the column 
* Pass the column specs to a `TableSpec` and add entries; The table spec is responsible for adding entry objects.
* Pass the table to the Panel Spec, and yield the panel :D 

In addition to this, 3 other important notes are:
1. A `Formatter` function can be given to a column spec, which will format the data (i.e. wrap the string)
2. One or many `Converter` functions can be given to a column spec. These would be used to alter the string in some way (i.e. `name = entry.name + " " + entry.short`. 
3. The `AbstractTableEntry` has a member `extras`. All members of `AbstractTableEntry.extras` will be treated as if they are members of `AbstractTableEntry`. This makes it easy for new members to be added to the entry, and later rendered if desired. This new members can  be callables (that return RenderableType), to allow for dynamic data to be shown. 

## Benefits of Changes... 
1. The layout, color, formatting, etc. of the Panel can now be more easily adjusted.  
2. Abilty to add / move / remove columns 
3. Ability add arbitary information / data to the panel as well as the ability to apply fun / interesting converters. 
4. More modular design should help the related issues mentioned above. 
5. Dynamic data provides some additionally flexibility. 

## Implementation Notes

1. `ColumnSpec.render_cell()` renders a single cell for a single entry. In order to render, the name of the entry member that is being rendered must match `ColumnSpec.key`. For example, to get the entry, `ColumnSpec.render_cell(entry)` runs `out = entry.get(self.key, None)`

Just something to keep in mind for users (if they are interfacing directly with ColumnSpec and AbstractTableEntry), or library development. 

## Other notes / Questions 

**Tests:** To date, I have only made sure the current unit-tests pass. New tests for custom Panel are needed. Currently all tests pass. 

**Question about Usage:** A way of leveraging these additionally would be beneficial. How will the user likely customized the panel? Via the arguments to `app.command(MyCustomHelpMembers=...)`? Or arguments to the app instance, or provided to the App class? 

**New Tests?** Once there exists some way to interface with these new abstracts in order to adjust the HelpPanel, then writing tests will likely be easier. Therefore, we may want to add those first.


ESPR3SS0:
Thats my bad... I'll be working on those

cyclopts/help/converters.py
```diff
@@ -0,0 +1,39 @@
...
+    from cyclopts.help import AbstractTableEntry
+
+
+def asterisk_converter(_: RenderableType, inp: AbstractTableEntry) -> RenderableType:
+    if inp.required:
+        return "*"
+    return ""
+
+
+def stretch_name_converter(_: RenderableType, inp: AbstractTableEntry) -> RenderableType:
```
BrianPugh: I'm a big worried by the existence of this. I haven't gotten to how it's used yet, but it would suggest that names are being combined into a single string somewhere else.

cyclopts/help/converters.py
```diff
@@ -0,0 +1,39 @@
...
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from rich.console import RenderableType
+
+    from cyclopts.help import AbstractTableEntry
+
+
+def asterisk_converter(_: RenderableType, inp: AbstractTableEntry) -> RenderableType:
```
BrianPugh: let's rename this `asterisk_required_converter` or something like that. People might be mislead that this has something to do with `Parameter(name="*")`, which is totally unrelated.

cyclopts/help/converters.py
```diff
@@ -0,0 +1,39 @@
+from __future__ import annotations
```
BrianPugh: lets remove the `from __future__ import annotations` if possible from the files. `from __future__ import annotations` actually has a lot of weird impacts on the type system, and I'd like to get by without it if possible. It's also going to be deprecated by 2029, if we want to be that forward thinking 🙈

cyclopts/help/help.py
```diff
@@ -64,29 +69,20 @@ def docstring_parse(doc: str, format: str):
...
 class InlineText:
-    def __init__(self, primary_renderable: "RenderableType", *, force_empty_end=False):
+    def __init__(self, primary_renderable: RenderableType, *, force_empty_end=False):
         self.primary_renderable = primary_renderable
         self.texts = []
         self.force_empty_end = force_empty_end
 
     @classmethod
-    def from_format(cls, content: Optional[str], format: str, *, force_empty_end=False):
+    def from_format(cls, content: str | None, format: str, *, force_empty_end=False):
```
BrianPugh: Unfortunately, we have to use `Union` until python3.9 hits EOL in... oh October 2025? That's just like 45 days away! Best to be python3.9 compatible though, and we'll drop python3.9 compatability in Cyclopts v4.

cyclopts/help/help.py
```diff
@@ -192,12 +188,145 @@ def __rich_console__(self, console, options):
             yield from wrapped_segments
 
 
+def _resolve(v: str | LazyData | None) -> str | None:
+    if v is None:
+        return None
```
BrianPugh: these 2 lines are technically not necessary, right?

cyclopts/help/help.py
```diff
@@ -192,12 +188,145 @@ def __rich_console__(self, console, options):
...
+    return v() if callable(v) else v
+
+
+# TODO: Is there a low cost runtime validator that all members in this
+#       are renderable?
```
BrianPugh, line 193: maybe something like:

```python
def __attrs_post_init__(self):
    for field in attrs.fields(type(self)):
        if not issubclass(field.type, rich.RenderableType):
            raise TypeError
```

but that would require an explicit import of Rich; but if users are customizing the help page, maybe that's unavoidable? Unless we had some sort of "help setup" hook to keep startup speed fast?

BrianPugh (commented):
I think this all looks great so far! I'll need to actually play around with it, but the code looks very well organized/written.

> Question about Usage: A way of leveraging these additionally would be beneficial. How will the user likely customized the panel? Via the arguments to app.command(MyCustomHelpMembers=...)? Or arguments to the app instance, or provided to the App class?

I think it would be both. As you've probably figured out by now, `@app.command` just creates another `App` object, so they're all one-in-the-same. We'd probably have to do some inheritence stuff. I would be hesitent to touch that at the moment because I plan on doing some revamping to that system soon for Cyclopts v4 (See #497).

cyclopts/help/converters.py
ESPR3SS0: In the original HelpPanel there were these lines (around line 268 in help.py): 

```python
...
            for entry in self.entries:
                name = entry.name
                if entry.short:
                    name += " " + entry.short
                name = "\n".join(wrap(name, commands_width))
                table.add_row(name, entry.description)
...
```

So the rendered name was f"{entry.name} {entry.short}". This was done _only_ for rendering command names. 

So yes they were combined into a single string, but just before they were rendered. The entry itself is _not_ altered, if that's what you were concerned about. 

With the new implementation, it would probably (maybe?), be "more correct" to have a ColumnSpec that is dedicate to printing `entry.short`, but I was trying to match the old logic as close as possible.

cyclopts/help/converters.py
ESPR3SS0: Sounds good to me, I'll pop that from all files.

cyclopts/help/help.py
ESPR3SS0: Gotcha, I have a whole lot of `foo | None` 😅, I'll convert those back to `Optional` and `Union`'s.

cyclopts/help/converters.py
BrianPugh: got it, i think your current implementation is actually totally reasonable.

cyclopts/help/help.py
ESPR3SS0: Ah yes, those are unnecessary!

cyclopts/help/converters.py
ESPR3SS0: This is changed now 👍

cyclopts/help/converters.py
ESPR3SS0: This is changed now as well 👍

cyclopts/help/help.py
ESPR3SS0: I was thinking something similar; I did not know how much of a priority startup time is.

My gut reaction was; if there is a non-renderable type, the trace back should be nice enough to make it clear that a renderable type is needed (but as I type the it sounds quite naive.). And thus the potential overhead is not worth it. 

Side note - on the topic of performance and out of my curiosity; are there benchmarks for performance / startup time? Even without the runtime checking mentioned above, I am curious how the dynamic rendering of the panel will affect the runtime.

ESPR3SS0:
I pushed changes for pretty much everything discussed above ( besides the runtime checking of members, that has no changes). The pre-commit passed, and tests all passed, so that side of things should all be good now 🤞. 

Next I will be playing with the new features more... and seeing if I can break anything.

cyclopts/help/protocols.py
```diff
@@ -0,0 +1,35 @@
+from typing import TYPE_CHECKING, Protocol, runtime_checkable
+
+if TYPE_CHECKING:
+    from rich.console import RenderableType
+
+    from .help import AbstractTableEntry
```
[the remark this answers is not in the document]
BrianPugh: invalid

cyclopts/help/help.py
```diff
@@ -13,17 +13,21 @@
...
 
-from attrs import define, field
+from attrs import Factory, define, evolve, field
 
 from cyclopts._convert import ITERABLE_TYPES
 from cyclopts.annotations import is_union, resolve_annotated
 from cyclopts.field_info import signature_parameters
 from cyclopts.group import Group
-from cyclopts.utils import SortHelper, frozen, resolve_callables
+from cyclopts.help.converters import asterisk_required_converter, combine_long_short_converter, stretch_name_converter
```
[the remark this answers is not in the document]
BrianPugh: invalid

cyclopts/help/formatters.py
```diff
@@ -0,0 +1,24 @@
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from rich.console import RenderableType
+
+    from cyclopts.help import AbstractTableEntry, ColumnSpec
```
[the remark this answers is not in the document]
BrianPugh: invalid

cyclopts/help/converters.py
```diff
@@ -0,0 +1,39 @@
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from rich.console import RenderableType
+
+    from cyclopts.help import AbstractTableEntry
```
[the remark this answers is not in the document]
BrianPugh: invalid

BrianPugh:
@ESPR3SS0 Copilot did a review; mostly noise/typos, but there is one comment about some commented code that might require a second look.

cyclopts/help/specs.py
```diff
@@ -0,0 +1,200 @@
...
+            no_wrap=self.no_wrap,
+        )
+
+    def render_cell(self, entry: "AbstractTableEntry") -> RenderableType:
+        """Render the cell."""
+        raw = entry.get(self.key, None)
+        out = raw(entry) if callable(raw) else raw
+        entry.try_put(self.key, out)
+
+        if self.converters:  # and out is not None:
```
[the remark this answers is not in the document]
ESPR3SS0: Ah yes; I switched to allowing a converter function to take an `out` that is `None`. 

This is okay because a converter takes the whole `entry` as input as well. For example, the converter `asterisk_required_converter` does not use the current value of `out`, so in that case `out` can be None. 

I'll remove that commented out portion.

BrianPugh:
@ESPR3SS0 what do we think the order of operations should be? Merge this in and then make another PR exposing some of the new functionality? Keep this as a PR and base a feature-branch off of it?

ESPR3SS0:
If you are feeling good about it, I would vote to Merge this in, then expose more functionality in a new PR. 

Given that Cyclopts v4 is going to have some revamped core logic,  I assume we'll wait until that is finalized to decide exactly how the new functionality will be exposed. 

Thus, getting this in now, then waiting for v4 changes, then exposing functionality sounds like a sensible route. Let me know what you think.

BrianPugh:
sounds good! I'll do one more review tonight and then merge this in. Because no external functionality has changed, there won't be a release, but it'll certainly be progress! Thanks again!

cyclopts/help/help.py
```diff
@@ -192,12 +188,143 @@ def __rich_console__(self, console, options):
...
+    justify="left",
+    style="cyan",
+    formatter=wrap_formatter,
+    converters=[stretch_name_converter, combine_long_short_converter],
+)
+
+
 @define
-class HelpPanel:
+class AbstractRichHelpPanel:
```
BrianPugh: Thoughts about revert the name of this back to just `HelpPanel`? It's not exactly "Abstract" since we instantiate it, and we're pretty deeply entrenched with Rich, so I'm not sure if we need that information either.

BrianPugh: (same goes for `AbstractTableEntry`. I think maybe these names made more sense in an earlier iteration).

cyclopts/help/help.py
```diff
@@ -192,12 +188,143 @@ def __rich_console__(self, console, options):
...
+    entries: list[AbstractTableEntry] = field(factory=list)
+
+    column_specs: list[ColumnSpec] = field(
+        default=Factory(
+            lambda self: [CommandColumn, DescriptionColumn]
+            if self.format == "command"
+            else [NameColumn, DescriptionColumn],
+            takes_self=True,
+        )
+    )
```
BrianPugh: I think that `column_specs` shouldn't exist here. In `__rich_console__`, the `self.table_spec.column_specs` gets overwritten with this value, which could definitely be a source of confusion. I think if the user wants to define a column_spec, that they do it through the `table_spec`.

cyclopts/help/help.py
```diff
@@ -192,12 +188,143 @@ def __rich_console__(self, console, options):
...
+    """Adjust the Format for the help panel!."""
+
+    from rich.box import ROUNDED, Box
+    from rich.console import Group as RichGroup
+    from rich.console import NewLine, RenderableType
+    from rich.panel import Panel
+    from rich.table import Column, Table
+    from rich.text import Text
+
     format: Literal["command", "parameter"]
```
BrianPugh: It was kind of a bodge before to have this `format` field, but I think you've organized the code enough now that it makes even less sense. I think it might make sense to have 2 subclasses here (`CommandHelpPanel` and `ParameterHelpPanel`). I think this also helps resolve the below "TODO".

cyclopts/help/help.py
```diff
@@ -225,84 +359,59 @@ def __rich_console__(self, console: "Console", options: "ConsoleOptions") -> "Re
...
-        wrap = partial(
-            textwrap.wrap,
-            subsequent_indent="  ",
-            break_on_hyphens=False,
-            tabsize=4,
-        )
-        # (top, right, bottom, left)
-        table = Table.grid(padding=(0, 2, 0, 0), pad_edge=False)
-        panel_description = self.description
+        commands_width = ceil(console.width * 0.35)
```
BrianPugh: these width calculations no longer belong here, right? I think this belongs in `ColumnSpec`?

cyclopts/help/help.py
ESPR3SS0: Sounds good to me, you're right these made sense early on but are now clunky.

cyclopts/help/help.py
ESPR3SS0: I agree, controlling the specs solely through the table_spec will be more clear.

ESPR3SS0:
Just wanted to mention I've been working up a change to address this... should have a little example I can paste here later today that we can discuss and make sure it looks good. Then maybe a commit or two later later today! 

It'll address removal of the old logic, and the Literal.

BrianPugh:
thanks for the update! and thank you again for your hard work!

ESPR3SS0:
Yeah no worries! 

Alright, so I've address the moving of the logic...ended up being alittle more than I expected.  I will now check out speeding everything up.

But first a TL;DR of the changes...

I felt that a builder pattern fit well here for 3 main reasons...
1. I was hesitant to use subclasses because the functionality is the same between a CommandHelpPanel and ParameterHelpPanel. The only difference is the default list of `column_specs`. 
2. I needed / wanted to move some of the logic out of the `__rich_console__` method (mentioned in one of the comments above). this was tricky because some of the `ColumnSpecs` used the `console` object that is passed to `__rich_console__`. 
3. **Importantly** something I did not see before is that a user had _no way_ of using the `console` object with the last implementation if they did want to define a custom list of `ColumnSpecs` (I.E. `NamedColumn(width=console.width)`). Now they can! 

So in summary, this ended up being very flexible, kept one flexible `HelpPanel` object, exposed a nice way to register "default" panels, and give the user the ability to use the `console` (and `options`) objects when defining their columns specs. 

In addition to this some new features / important changes are... 
1. An in memory registry of preset panels. This (could, maybe, hopefully) be a convient way to expose functionality. The idea is to allow the usage of "default" column specs, but still allow the user to directly override `TableSpecs.columns`
2. Moved _all_ responsibility of handling the list of `ColumnSpecs` to the `TableSpecs` object. 
3. Not a feature... but satisfying... the `HelpPanel.__rich_console__` function is now _very very_ dumb (or simple). It (1) Sets the Description (2) Realizes columns, builds the table, and adds entry (3) yields the panel.

I didn't think this would be as large of a change, so thanks for bearing with all the code!

Again, let me know what you think about the logic.

cyclopts/help/converters.py
BrianPugh, line 1: is the input parameter `out` for each of these converters required? None of them currently use it. Is there planned future functionality that would require this object?

cyclopts/help/formatters.py
BrianPugh, line 1: commenting to revisit the formatter function signatures as well (same logic as cyclopts/help/converters.py)

cyclopts/help/help.py
```diff
@@ -192,12 +185,117 @@ def __rich_console__(self, console, options):
...
+    extras: TableData = field(factory=TableData, repr=False)
+
+    def try_put(self, key: str, value: Optional[Union["RenderableType", "LazyData"]]):
+        """Put a attr to the dataclass.
+
+        This is looser than put, and will not raise an AttributeError if
+        the member does not exist. This is useful when the list of entries
+        do not have all the same members. This was required for
+        `ColeSpec.render_cell`
+        """
```
BrianPugh, line 231: to have less duplicate logic, it might be better to implement this like:
```python
try:
    return self.put(key, value)
except AttributeError:
    return self
```

cyclopts/help/specs.py
```diff
@@ -0,0 +1,299 @@
...
+    from rich.console import RenderableType
+    from rich.style import Style
+    from rich.table import Table
+
+    PaddingType = Union[int, tuple[int, int], tuple[int, int, int, int]]
+
+    key: str
+
+    formatter: Optional["Formatter"] = None
+    converters: Optional[Union["Converter", list["Converter"]]] = None
```
BrianPugh: to make this consistent with other classes in cyclopts, use `field(converter=to_tuple_converter)` so that it will always be a `tuple[Converter, ...]`, allowing all downstream code to always iterate over them.

cyclopts/help/specs.py
```diff
@@ -0,0 +1,299 @@
...
+from cyclopts.help.formatters import wrap_formatter
+
+if TYPE_CHECKING:
+    from rich.console import Console, ConsoleOptions
+
+    from cyclopts.help import TableEntry
+    from cyclopts.help.protocols import ColumnSpecBuilder, Converter, Formatter
+
+
+@define(frozen=True)
```
BrianPugh: nit: use the `@frozen`  from `from cyclopts.utils import frozen`. It's a small tweak that reduces object construction cost, thus speeding up import-time a tiny bit. Same thing for other uses of `@define(frozen=True)` in this PR.

cyclopts/help/specs.py
```diff
@@ -0,0 +1,299 @@
...
+        ]
+
+
+_COLUMN_SPEC_BUILDERS: dict[str, "ColumnSpecBuilder"] = {
+    "command": _command_column_spec_builder,
+    "parameter": _parameter_column_spec_builder,
+}
+
+
+def register_panel_columnbuilder(name: str, builder: "ColumnSpecBuilder", *, overwrite: bool = False) -> None:
```
BrianPugh, line 159: I know that this isn't called anywhere, but I worry about anything that mutates a global object in Cyclopts. All mutable state should be confined to the user-instantiated `App` object. Multiple libraries could call `register_panel_columnbuilder`, thinking that they have exclusive access and cause a whole bunch of headache.

cyclopts/help/specs.py
```diff
@@ -0,0 +1,299 @@
...
+        """Insert the entries into the table."""
+        if callable(self.columns):
+            raise TypeError("Columns must be realized before building table")
+
+        for e in entries:
+            cells = [col.render_cell(e) for col in self.columns]
+            table.add_row(*cells)
+
+    # To help with padding...
+    def with_padding(self, padding: PaddingType) -> "TableSpec":
```
BrianPugh: any benefits of an explicit method over the equally verbose `obj.with_(padding=42)` (which I like, very clean!)? Same comments for the other specs.

cyclopts/help/converters.py
ESPR3SS0: Good point!

I included `out` before added `try_put`, and thus using `out` was the only way to use the updated input. This could be removed now.

BrianPugh:
skimming over this, it's looking good! our main blocker right now is the import time, correct?

ESPR3SS0:
Yes that and the global register for column builders. 

Life is picking up again... making things a wee bit busier but I hope to address both of these here soon. 

The current ideas for fixes: 
- Move the register to the App instance 
-  Move 'CyclpoptePanel' to it's own file and lazily import

BrianPugh:
i might try and take a stab at it; i'll try working on it (in a separate branch) soon!

BrianPugh:
this has been merged into the `v4-develop` branch! We can play with it more over there; i expect to do a v4 release within the next 2-4 weeks.

Issue #520, opened 2025-08-21
Hidden commands are displayed in unknown command error messages


ryan-thom:
A hidden command (`show=False`) is included in the listing of available commands when an unknown command error message is displayed.
```python
from cyclopts import App

app = App()


@app.command
def foo():
    print("Running foo.")


@app.command(show=False)
def bar():
    print("Running bar.")


if __name__ == "__main__":
    app()
```
When executed with an unknown command as input, the output includes and references the hidden `bar` command:
```shell
> python demo.py barr
╭─ Error ───────────────────────────────────────────────────────────────────────────────────────────╮
│ Unknown command "barr". Did you mean "bar"? Available commands: foo, bar.                         │
╰───────────────────────────────────────────────────────────────────────────────────────────────────╯
```

This was not the behavior I would have expected, however, I can see that this may be intended/wanted as the command is executable and is therefore "available". If it is intended, it might be worth mentioning this behavior in the API docs

BrianPugh:
I hadn't really thought about it before, but I think the current behavior is **sort of** fine/intended. The [API docs](https://cyclopts.readthedocs.io/en/latest/api.html#cyclopts.App.show) state that `App.show` toggles whether or not the command is shown **on the help screen**.

However, when displaying this "did you mean" screen, we do print the commands (if there are less than 8 commands), as you showed. And I imagine that if the developer doesn't want the command to be shown on the help-page, they probably don't want it shown on this page either. So I can change this "did you mean" screen to respect the `show` flags. Do you think that would be a good enhancement?

ryan-thom:
I think that would be a good enhancement. At least for me, it would better align behavior with my natural expectation of what a "hidden" option achieves. As you mentioned, my expectation was stemming from the same idea that if I don't want the command displayed on the help-page, then I don't really want it to be shown/recommended/discoverable through a means I can't control.

BrianPugh:
alright sounds good! I'll make a PR for this soon!

BrianPugh:
resolved in v3.23.0

PR #521, opened 2025-08-23
Don't recommend commands that have `show=False`.

BrianPugh:
Addresses #520. @ryan-thom what do you think of this?


ryan-thom:
That's perfect! Thank you!

Issue #519, opened 2025-08-20
"--foo" requires an argument, even though --foo is a dataclass with no required members


andersjel:
For the below code:
```python
from dataclasses import dataclass
import cyclopts

@dataclass
class Foo:
    bar: int = 12

def cmd(foo: Foo):
    print(foo)

if __name__ == "__main__":
    cyclopts.run(cmd)
```
I expect:
```shell
> python example.py
Foo(bar=12)
```
Instead I get:
```plain
> python example.py
╭─ Error ──────────────────────────────────────────────────╮
│ Parameter "--foo" requires an argument.                  │
╰──────────────────────────────────────────────────────────╯
```

Note:
```plain
> python example.py --help
Usage: cmd [ARGS] [OPTIONS]

╭─ Commands ───────────────────────────────────────────────╮
│ --help -h  Display this message and exit.                │
│ --version  Display application version.                  │
╰──────────────────────────────────────────────────────────╯
╭─ Parameters ─────────────────────────────────────────────╮
│ FOO.BAR --foo.bar  [default: 12]                         │
╰──────────────────────────────────────────────────────────╯
```

As a workaround, I can modify the definition of _cmd_ to read `def cmd(foo: Foo = Foo()): ...`.

BrianPugh:
Hey @andersjel!

At first glance, this seems like a totally reasonable expectation (and I even began implementing it at one point)! But underneath the surface, it contradicts a lot of Cyclopts logic.

Consider just the function signature:

```python
def cmd(foo: Foo):
    print(foo)
```

From interpreting this signature, we can see that `foo` is a required parameter; it clearly has no default value. If we had the following signature:

```python
def cmd(foo: int):
    print(foo)
```

We would certainly not expect `foo` to default to `int() == 0`. It's a required parameter!

So, in your case, if `Foo` is immutable, then I would consider your workaround as the proper solution. If it is not, then the alternative solution is:

```python
def cmd(foo: Foo | None = None):
    if foo is None:
        foo = Foo()
    print(foo)
```

A benefit of this is that it is standard python that most developers would be familiar with.

andersjel:
Thank you for the quick response.

I saw this initially with code like

```python
@dataclass
@cyclopts.Parameter(name='*')
class Foo:
    bar: int = 12
```

which I think makes it a bit more surprising. You end up with:

```plain
❯ python example.py --help
Usage: cmd [ARGS] [OPTIONS]

[...]

PR #523, opened 2025-08-23
Generate an error message if a Parameter(name="*") parameter has all optional values, but itself does not have a default value.

BrianPugh:
Addresses #519 .

@andersjel can you play with this and see if it behaves as expected?

Marked currently as a draft because this will have to go into v4 instead of v3 due to it likely breaking someone's code.

Minimal Example:
```python
import cyclopts

@dataclass
@cyclopts.Parameter(name="*")
class Foo:
    bar: int = 12

def main(foo: Foo):
    pass

cyclopts.run(main)
```

### DIFF
--- a/cyclopts/parameter.py
+++ b/cyclopts/parameter.py
@@ -430,13 +430,14 @@ def validate_command(f: Callable):
 
             if all_fields_optional:
                 param_name = field_info.names[0] if field_info.names else ""
-                if param_name:
-                    param_name = f'"{param_name}" '
+                quoted_param_name = f'"{param_name}" ' if param_name else ""
                 raise ValueError(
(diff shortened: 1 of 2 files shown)


andersjel:
> @andersjel can you play with this and see if it behaves as expected?

Sure, I tried running from the branch and it gives an exception as expected.

One minor nitpick, the output looks like:

```
... Consider either:
    1) providing a default value like ""args" : Args = Args()"
    2) making it optional like ""args" : Args | None = None".
```

but should it not be

```
... Consider either:
    1) providing a default value like "args: Args = Args()"
    2) making it optional like "args: Args | None = None".
```

without the extra quotes and space around args?

BrianPugh:
good catch! yes i'll fix that

BrianPugh:
This has been merged into the `v4-develop` branch and will be a part of the v4 release.

PR #530, opened 2025-08-30
`negative_iterable` with an empty string shouldn't appear in the help screen

### LINKED ISSUE #529: [BUG] `negative_iterable=""` appears in the help screen
I'm not sure if this is a bug or the intended behavior. 

Here's a reproducible example:

```python
# /// script
# dependencies = [
#   "cyclopts",
# ]
#  requires-python = ">=3.12"
# ///

from typing import Annotated

app = App(default_parameter=Parameter(negative_iterable=""))


@app.command()
def fetch(
    urls: Annotated[list[str] | None, Parameter(alias="-u", consume_multiple=True)] = None,
    cache: Annotated[bool, Parameter(alias="-c")] = True,
):
    """
    Example fetch.

    :param urls: A list of URLs
    :param cache: Whether to cache
    """
    pass


if __name__ == "__main__":
    app()
```

I would expect the output to be:

```
Example fetch.

╭─ Parameters ────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ URLS --urls               -u  A list of URLs                                                                        │
│ CACHE --cache --no-cache  -c  Whether to cache [default: True]                                                      │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

But it's actually this:

```
Example fetch.

╭─ Parameters ────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ URLS --urls --urls        -u  A list of URLs                                                                        │
│ CACHE --cache --no-cache  -c  Whether to cache [default: True]                                                      │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

I know I can get my expected output by setting `negative=""` in `urls`, but if I have multiple list parameters, I have to set it in every single one. If I put `negative=""` in the `App` instead, I'd have to manually put `negative="no-"` on every boolean parameter.

nachocab:
Fixes #529

### DIFF
--- a/cyclopts/parameter.py
+++ b/cyclopts/parameter.py
@@ -291,7 +291,8 @@ class Parameter:
             assert isinstance(negative_prefixes, tuple)
             if self.negative is None:
                 for negative_prefix in negative_prefixes:
-                    out.append(f"--{name_prefix}{negative_prefix}{name_components[-1]}")
+                    if negative_prefix:
+                        out.append(f"--{name_prefix}{negative_prefix}{name_components[-1]}")
             else:
                 for negative in user_negatives:
                     out.append(f"--{name_prefix}{negative}")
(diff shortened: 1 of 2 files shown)


BrianPugh:
Thank you! I will release this right now under v3.23.1!

PR #531, opened 2025-08-30
Improve contributing docs and .gitignore

nachocab:
It took me a while to realize that `pytest` wasn't the right way to run the test suite (it resulted in 19 failing tests), so I've updated the documentation.

Also, it would be useful to gitignore the .vscode folder so I can disable "Auto-format on Save" for this repo.

### DIFF
--- a/.github/contributing.md
+++ b/.github/contributing.md
@@ -19,13 +19,14 @@ Good docstrings include information like:
 3. When dealing with physical values, include units.
 
 ## Unit Tests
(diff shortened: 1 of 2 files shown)


BrianPugh:
Thank you! Coincidentally, i just recently [made the vanilla `pytest` command work](https://github.com/BrianPugh/cyclopts/commit/d47fd4de7cb4b970b0abfceaa61d9e1ac82b30f4) on the [v4-develop branch](https://github.com/BrianPugh/cyclopts/tree/v4-develop). Regardless, this is good information to document, thank you!

Issue #532, opened 2025-09-01
Using async function as meta app entrypoint throws RuntimeError


schaermu:
I'm trying to do some async initialization work (GQL client) before running the actual cyclopts app:

```python
# cli/__main__.py
import sys
from typing import Annotated
from cyclopts import App, Parameter
from rich.console import Console

from . import github_client
from .measure import app as measure_app

console = Console()
app = App(console=console)

app.command(measure_app, name="measure")

@app.meta.default
async def launcher(*tokens: Annotated[str, Parameter(show=False, allow_leading_hyphen=True)]):
    try:
        await github_client.connect()
        app(tokens)
    except Exception:
        console.print_exception()
        sys.exit(1)
    finally:
        await github_client.close()

if __name__ == "__main__":
    app.meta()
```

Some notes:
- `measure_app` is a dynamically created app consisting of multiple commands being read from the directory. `pr_cycle_time` is the command being invoked in this example.
- The sub-commands access the initialized `github_client` when querying GitHub API.

When running this implementation using `uv run -m cli measure pr-cycle-time <owner>/<repo>`, i'm getting the following error:

```python
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /home/schaermelist/projects/post/schaermu-devx-measure-app/cli/__main__.py:22 in launcher        │
│                                                                                                  │
│   19 async def launcher(*tokens: Annotated[str, Parameter(show=False, allow_leading_hyphen=Tr    │
│   20 │   try:                                                                                    │
│   21 │   │   await github_client.connect()                                                       │
│ ❱ 22 │   │   app(tokens)                                                                         │
│   23 │   except Exception:                                                                       │
│   24 │   │   console.print_exception()                                                           │
│   25 │   │   sys.exit(1)                                                                         │
│                                                                                                  │
│ /home/schaermelist/projects/post/schaermu-devx-measure-app/.venv/lib/python3.10/site-packages/cy │
│ clopts/core.py:1285 in __call__                                                                  │
│                                                                                                  │
│   1282 │   │   │   │   if backend == "asyncio":                                                  │
│   1283 │   │   │   │   │   import asyncio                                                        │
│   1284 │   │   │   │   │                                                                         │
│ ❱ 1285 │   │   │   │   │   return asyncio.run(command(*bound.args, **bound.kwargs))              │
│   1286 │   │   │   │   elif backend == "trio":                                                   │
[...]

PR #535, opened 2025-09-01
detect if we're in a async context, and if so return the coroutine instead of running it.

BrianPugh:
In response to #532 .

@schaermu I don't regularly use async, so let me know what you think of this solution. Does it seem reasonable? Does it seem intuitive? I'm a little unsure of performing different actions depending on if we're in an async context or not, so please let me know your thoughts!

### DIFF
--- a/cyclopts/core.py
+++ b/cyclopts/core.py
@@ -1211,6 +1211,47 @@ class App:
 
         return command, bound, ignored
 
+    def _run_with_asyncio(self, command: Callable, bound: inspect.BoundArguments):
+        """Run async command with asyncio backend."""
+        import asyncio
+
+        try:
+            # Check if we're already in an async context
+            asyncio.get_running_loop()
+        except RuntimeError:
+            # No running loop, create one
+            return asyncio.run(command(*bound.args, **bound.kwargs))
+        else:
+            # We're in an async context, return the coroutine for the caller to await
+            return command(*bound.args, **bound.kwargs)
+
+    def _run_with_trio(self, command: Callable, bound: inspect.BoundArguments):
+        """Run async command with trio backend."""
+        from functools import partial
+
+        import trio
+
+        try:
+            # Check if we're already in a trio context
+            trio.lowlevel.current_trio_token()
+        except RuntimeError:
+            # No trio context, create one
+            return trio.run(partial(command, *bound.args, **bound.kwargs))
+        else:
+            # We're in a trio context, return the coroutine for the caller to await
+            return command(*bound.args, **bound.kwargs)
+
+    def _execute_async_command(
+        self, command: Callable, bound: inspect.BoundArguments, backend: Literal["asyncio", "trio"]
+    ):
+        """Execute an async command with the specified backend."""
+        if backend == "asyncio":
+            return self._run_with_asyncio(command, bound)
+        elif backend == "trio":
+            return self._run_with_trio(command, bound)
+        else:  # pragma: no cover
+            assert_never(backend)
+
     def __call__(
         self,
         tokens: Union[None, str, Iterable[str]] = None,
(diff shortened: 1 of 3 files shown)


gerlero:
Sorry for the delay (I said in #536 I'd take a look at this one)!

I'm looking at it right now and will leave my comments with the code above.

cyclopts/core.py
```diff
@@ -1282,11 +1282,25 @@ def __call__(
...
 
-                    return asyncio.run(command(*bound.args, **bound.kwargs))
+                    try:
+                        # Check if we're already in an async context
+                        asyncio.get_running_loop()
+                        # We're in an async context, return the coroutine for the caller to await
+                        return command(*bound.args, **bound.kwargs)
+                    except RuntimeError:
+                        # No running loop, create one
+                        return asyncio.run(command(*bound.args, **bound.kwargs))
```
gerlero:
```suggestion
                    try:
                        # Check if we're already in an async context
                        asyncio.get_running_loop()
                    except RuntimeError:
                        # No running loop, create one
                        return asyncio.run(command(*bound.args, **bound.kwargs))
                    else:
                        # We're in an async context, return the coroutine for the caller to await
                        return command(*bound.args, **bound.kwargs)
```

to avoid swallowing any unrelated (also unexpected, I give it to you) `RuntimeError`s from instantiation of the coroutine

cyclopts/core.py
```diff
@@ -1282,11 +1282,25 @@ def __call__(
                 if backend == "asyncio":
                     import asyncio
 
-                    return asyncio.run(command(*bound.args, **bound.kwargs))
+                    try:
+                        # Check if we're already in an async context
+                        asyncio.get_running_loop()
+                        # We're in an async context, return the coroutine for the caller to await
+                        return command(*bound.args, **bound.kwargs)
```
gerlero: Is there a chance a caller might miss the fact that a coroutine was returned and needs to be awaited? Asking because it's looks like an interesting pattern but I'm not sure what part of the code would have to deal with this return.

cyclopts/core.py
```diff
@@ -1282,11 +1282,25 @@ def __call__(
...
 
-                    return trio.run(partial(command, *bound.args, **bound.kwargs))
+                    try:
+                        # Check if we're already in a trio context
+                        trio.lowlevel.current_trio_token()
+                        # We're in a trio context, return the coroutine for the caller to await
+                        return command(*bound.args, **bound.kwargs)
+                    except RuntimeError:
+                        # No trio context, create one
+                        return trio.run(partial(command, *bound.args, **bound.kwargs))
```
gerlero:
```suggestion
                    try:
                        # Check if we're already in a trio context
                        trio.lowlevel.current_trio_token()
                    except RuntimeError:
                        # No trio context, create one
                        return trio.run(partial(command, *bound.args, **bound.kwargs))
                    else:
                        # We're in a trio context, return the coroutine for the caller to await
                        return command(*bound.args, **bound.kwargs)
```

gerlero: Maybe we can abstract all of this logic into a utility function? TBH it would be really useful for implementing #536

cyclopts/core.py
BrianPugh: I don't think there is, but this whole "sometimes return a coroutine, sometimes execute it" makes me a little uncomfortable. But, it could also be the "it just works" solution.

cyclopts/core.py
BrianPugh: yup exactly! Once we determine that this PR is something we actually want to do, we can abstract out the common parts, update #536, and then merge that too.

BrianPugh:
@schaermu what are your thoughts?

cyclopts/core.py
BrianPugh: (with upcoming v4, i'm comfortable with any breaking changes that make it more ergonomic/intuitive/consistent)

BrianPugh:
With some hesitation, I'm going to merge this. I think it's better than the current situation, and the pragmatic downsides might not actually really occur in people's real code.

gerlero:
Sorry for not checking back, although there's not much more input I could've given here. I'm happy that this is merged and will update #536 with the changes here.

PR #536, opened 2025-09-01
Allow `asyncio` coroutines as callable `App` `version` arguments

gerlero:
Adds support for passing an asyncio-based coroutine to the `version` argument of `Apps`. Also adds a test to check that this works.

### DIFF
--- a/cyclopts/core.py
+++ b/cyclopts/core.py
@@ -238,7 +238,7 @@ class App:
         kw_only=True,
     )
 
-    version: Union[None, str, Callable[..., str]] = field(default=None, kw_only=True)
+    version: Union[None, str, Callable[..., str], Callable[..., Coroutine[Any, Any, str]]] = field(default=None, kw_only=True)
     # This can ONLY ever be a Tuple[str, ...]
     _version_flags: Union[str, Iterable[str]] = field(
         default=["--version"],
@@ -554,7 +554,15 @@ class App:
                 except PackageNotFoundError:
                     pass
         else:
-            version_raw = self.version() if callable(self.version) else self.version
+            if callable(self.version):
+                if inspect.iscoroutinefunction(self.version):
+                    # For async version callables, use asyncio as the default backend
+                    import asyncio
+                    version_raw = asyncio.run(self.version())
+                else:
+                    version_raw = self.version()
+            else:
+                version_raw = self.version
 
         if not version_raw:
             version_raw = _default_version()
--- a/tests/test_version.py
+++ b/tests/test_version.py
@@ -65,3 +65,16 @@ def test_version_print_help_format_override(app, console):
         app.version_print(console)
 
     assert "foo\n" == capture.get()
+
+
+def test_version_print_custom_async_callable(app, console):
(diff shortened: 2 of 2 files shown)


BrianPugh:
Hey @gerlero! Thanks for making this PR! Some of the logic is kind of related to the newly opened #535 . I was wondering if you could weight in over there? Depending on what we go with, we should maybe abstract out this coroutine detection/execution logic.

Otherwise I think this PR looks great and thank you for making it!

gerlero:
@BrianPugh thanks! I’ll take a look.

BrianPugh:
I think we might want to put this into v4 for the following reason:
1. Currently in `App.__call__`, we can specify the field `backend` to use different asynchronous libraries.
2. It is assumed that if someone wants to use something like `trio`, that they would also want to use it for the specified version coroutine.
3. Therefore we should add a `App.backend` attribute.
4. `v4-develop` has a better inheritance mechanism that makes this easy. All the other parameters from `App.__call__` like `exit_on_error` are available as `App` attributes and use the new inheritance machinery.

gerlero:
> I think we might want to put this into v4 for the following reason:
> 
> 1. Currently in `App.__call__`, we can specify the field `backend` to use different asynchronous libraries.
> 2. It is assumed that if someone wants to use something like `trio`, that they would also want to use it for the specified version coroutine.
> 3. Therefore we should add a `App.backend` attribute.

Yeah, I tend to agree.

> 4. `v4-develop` has a better inheritance mechanism that makes this easy. All the other parameters from `App.__call__` like `exit_on_error` are available as `App` attributes and use the new inheritance machinery.

Should I rebase on `v4-develop` (maybe also merge the changes of #535 there) or keep working on top of `main`?

BrianPugh:
so there's a bit of a conundrum now that I don't know how to solve:

1. With #535 merged in, I briefly attempted to integrate the logic with this PR.
2. It seems like fundamentally this won't work if we call `app` within a running async event loop. E.g.
   ```python
   def test_version_print_custom_async_callable_in_context(app, console):
    """Test that async callables work for version."""

    async def my_async_version():
        return "**async-foo**"

    app.version = my_async_version

    async def runner():
        with console.capture() as capture:
            app.version_print(console)
        return capture.get()

    result = asyncio.run(runner())
    assert "async-foo\n" == result
    ```
3. I think this could happen if the user is using an async meta app (original motivation behind #535).

So we need to figure out a solution, and if we cannot, maybe revert #535.

> Should I rebase on v4-develop (maybe also merge the changes of https://github.com/BrianPugh/cyclopts/pull/535 there) or keep working on top of main?

Lets focus on the current `main` base; I'll rebase, update it to v4-develop later.

BrianPugh:
I will play around with this PR with #539 and see if it all integrated together better.

gerlero:
Thanks! Feel free to rebase and modify this PR.

I have to work on a different project today, but do ping me if you have any questions or comments. BTW, thanks for developing this package!

BrianPugh:
closing in favor of #540 (that contains this PR)

PR #540, opened 2025-09-08
Allow async `App.version` handler

BrianPugh:
Built on-top of #539 and #536.

Performing an `asyncio.run` within `App.version_print` has 2 issues:

1. If we are already in an event loop, this would raise a `RuntimeError` (same basic issue as #532, which is why this is built ontop of #539).

2. If the user wants a different async backend (e.g. `trio`), then we're using the wrong event-runner.

So in this solution, we define an async method `App._version_print_async` (currently private because I don't think there's a strong benefit in making it public for now). The only difference between this and `App.version_print` is that it's async and will `await` a callable `self.version`. All of this is hidden to the user and will happen naturally through the mechanism introduced in #539.

This PR highlights a bug/issue that overriding the `App.default_command` for the `help_flags` (i.e. `app["--help"].default_command = my_custom_handler`) is not respected. We just directly use App.version_print (or `App._version_print_async` now) in `App._parse_args`. We can solve this in a future PR since it's not a newly introduced issue.

@gerlero can you review this (and play around with it). It should all work since all the new tests pass, but would love a second set of eyes on it. Assuming we think this is all good, we'll merge #539, then this (and close #536 as unmerged). Don't worry, i'll give you credit in the release notes 😉.


cyclopts/core.py
```diff
@@ -580,6 +583,39 @@ def _registered_commands(self) -> dict[str, "App"]:
...
+
+        Returns
+        -------
+        Optional[str]
+            Version string if found, None otherwise.
+        """
+        if self._instantiating_module is not None:
+            full_module_name = self._instantiating_module.__name__
+            root_module_name = full_module_name.split(".")[0]
+            try:
```
gerlero, line 576: @BrianPugh Should probably look for `__version__` attribute in the module too (like the `_default_version` does). 

If you ask me, a `__version__` attribute should take precedence over whatever `importlib.metadata` returns, but that's just my opinion (and it may be a breaking change for users).

cyclopts/core.py
```diff
@@ -580,6 +583,39 @@ def _registered_commands(self) -> dict[str, "App"]:
...
+                return importlib_metadata_version(root_module_name)
```
gerlero, line 577: No need for the conditional import and alias at the top: `importlib.metadata` already comes built-in with Python 3.8 (not 3.10 as the conditional import implies)

gerlero: If you get rid of these conditional imports, you should also remove the conditional dependency on  the `importlib_metadata` backport package.

cyclopts/core.py
BrianPugh: given that the `importlib` method is [more proper](https://packaging.python.org/en/latest/discussions/versioning/#accessing-version-information-at-runtime), we give it higher priority.

gerlero: Fair enough (I guess if `__version__` and `importlib` return different things for a package, you're either actively developing the package or the package author messed up)

cyclopts/core.py
BrianPugh: good catch! will update that here, thanks!

BrianPugh:
I made the changes and merged this in. @gerlero it would be great if you could give this one more test (in `main`). I'll tag a release later today if all is well!

gerlero:
> I made the changes and merged this in. @gerlero it would be great if you could give this one more test (in `main`). I'll tag a release later today if all is well!

It's working great for me! Thanks!

gerlero:
@BrianPugh Thanks! With this new release I can finally get rid of [my boilerplate `async_to_sync` decorator](https://github.com/gerlero/styro/blob/ca85c0bb72f349f50b0a7016c49bfa0db436d3e6/styro/_util.py#L19-L28). Happy to have switched to `cyclopts`!

PR #545, opened 2025-09-10
Expose help customization via groups.

BrianPugh:
Tagging @ESPR3SS0. This is beginning to expose your new refactored help stuff to cyclopts-users. I'll still need to play around with it more and add documentation/tests.

### DIFF
--- a/cyclopts/core.py
+++ b/cyclopts/core.py
@@ -1612,10 +1612,12 @@ class App:
                 try:
                     _, command_panel = panels[group.name]
                 except KeyError:
-                    command_panel = HelpPanel(
-                        format="command",
-                        title=group.name,
-                    )
+                    kwargs = {}
+                    if group.table_spec is not None:
+                        kwargs["table_spec"] = group.table_spec
+                    if group.panel_spec is not None:
+                        kwargs["panel_spec"] = group.panel_spec
+                    command_panel = HelpPanel(title=group.name, format="command", **kwargs)
                     panels[group.name] = (group, command_panel)
 
                 if group.help:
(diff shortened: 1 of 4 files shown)


BrianPugh:
@ESPR3SS0 heads up, in another branch, `v4-develop-help-customization-more-abstraction`, i'm modifying this code a lot. That branch will probably be merged into this branch (`v4-develop-help-customization`) this weekend.

High level changes:

1. There is now a higher level "formatter" interface who's job is to render HelpPanels. This allows for us to have significantly different help-screens (e.g. I'm working on a "plain" help screen described in #346 ).
2. I'm removing the generic `TableData` class (and the associated `TableEntry.extras` field). I understand the original purpose was to allow for the inclusion of arbitrary data, but I feel like it adds a lot of complexity. At the end of the day, the `TableEntry` object creation isn't really exposed to the cyclopts-user in any reasonably-friendly way. Instead, I'm attaching more generically-useful (but currently unused) fields to `TableEntry`, like `TableEntry.type`.  I'll also be attaching the associated `Argument`, from which basically any possible additional information could be derived, so that should cover our bases here.
3. I've combined `ColumnSpec`'s `formatters` and `converters` into a single `renderer` function. This started with me first updating `TableEntry.name: str   ->   TableEntry.names: tuple[str, ...]` so that we don't have to unnecessarily deal with the concatenated names. But, this also highlighted that the current `converters` system can unexpectedly write incorrect types to different attributes, which then made them tightly coupled with `formatters`. This coupling couldn't really be broken, so I just combined them. The de-coupling also made less sense as the converters became simpler due to better upstream data organization.
4.  I've got rid of the global `TableSpec`-registry and replaced them with `TableSpec` classmethods for generating the common Parameter and Parameter tables. The reasoning was explained before, but basically I don't want mutable global state within Cyclopts, as it could have hard-to-debug negative interactions between separate libraries using Cyclopts. The registry is also unnecessary because the cyclopts-user now has direct access to instantiating `TableSpec` and have it impact the help-page.

I might make a few more structural changes, but I think we're converging on a good, generic solution here.

BrianPugh:
tons of updates; i think it's mostly polishing docs and cleaning up tests now.

Issue #324, opened 2025-02-08
v4 Breaking Change Wishlist


BrianPugh:
This issue serves as a "Breaking Change" wishlist for Cyclopts. Please feel free to comment about thoughts/opinions presented here, as well as if you have any **breaking change** suggestions.

**Do not** use this thread as a generic feature wishlist. If a feature is backwards compatible (e.g. shell autocomplete), create a separate github issue for that.

v4 currently does not have a release date, but if this list gets sufficiently large (or a feature provides sufficient value), we can make this more concrete.

Features that have been implemented in `v4-develop` branch:

- [ ] Change `Parameter` arguments to `kw_only`

### Change `Parameter` arguments to `kw_only`
All arguments to `Parameter` really should be `kw_only`, with the possible exception of `name`, as it's a natural first positional argument. I suspect people always use all parameters (again, except for `name`) as `kw_only`, but it would be nice to formalize/codify it.

flying-sheep:
Hi! Thanks for the great library! I would like to add an argument against defaulting to Markdown: we’ve been extensively using MyST for our community’s packages, and had a mixed experience with it. On the con side, it mostly adds more complexity, potential for bugs, and another maintenance bottleneck on top of the already super complex Sphinx.

Of course you could also support MkDocs’ Markdown instead, but here’s the main point of my argument: Which Markdown would be the default?

There is only one rST, as it’s extensible by design with roles and directives. There are arbitrarily many Markdowns. So I’d say it’s smart to let users specify exactly which Markdown they want if they want it.

BrianPugh:
Thanks @flying-sheep for the thoughts/feedback! Currently Cyclopts just pipes the appropriate docstrings along to rich [Markdown](https://rich.readthedocs.io/en/stable/markdown.html) or rich [RestructuredText](https://github.com/wasi-master/rich-rst). The RestructuredText library currently works, but can be a bit buggy sometimes.

> So I’d say it’s smart to let users specify exactly which Markdown they want if they want it.

I believe that rich internally uses [markdown-it-py](https://github.com/executablebooks/markdown-it-py) for parsing the markdown; I think it can be tweaked for different styles (e.g. `gfm`). I'd have to dig a bit deeper to see if it's possible to expose various flavors.

> Another option would be no default

[...]

PR #569, opened 2025-09-30
V4: Add custom handling for sphinx rst directives.

BrianPugh:
@beroda can you give this a try? Note that v4 has many changes, but is mostly backwards compatible. See WIP Changelog in [this post](https://github.com/BrianPugh/cyclopts/issues/324#issue-2840084124).


edabor:
Great @BrianPugh ! Definitely better indeed.

image

PR #575, opened 2025-10-04
V4: import/happy-path  speed optimizations

BrianPugh:
@daudef can you give this branch a try to ensure that it doesn't negatively impact version detection? It has all the v4 changes (described at the bottom of [this post](https://github.com/BrianPugh/cyclopts/issues/324)); hopefully your application works with little to no modifications. This defers the (relatively) expensive `inspect.getmodule` call to later from #217 . This saves a few dozen mS.


seowalex:
Hi @BrianPugh, I think this breaks type checking (at least in Pyright) since the types cannot be statically determined. I think type stubs need to be added?

BrianPugh:
I believe you are right! I'll tag you in a PR momentarily to see if it fixes it.

BrianPugh:
https://github.com/BrianPugh/cyclopts/pull/604

Issue #572, opened 2025-10-01
Unable to create optional parameter that works both as flag and with multiple arguments


ilana-n:
Hello! I appreciate this library's functionality as a whole and for being able to support Pydantic - thanks in advance for giving this a look. 

For the below code:

```python
from typing import Annotated
from pydantic import BaseModel, Field
import cyclopts

class Config(BaseModel):
    urls: Annotated[
        list[str] | None,
        Field(
            default=None,
            description="Optional list of URLs"
        ),
        cyclopts.Parameter(
            consume_multiple=True,
        ),
    ]

def cmd(config: Config):
    print(f"urls: {config.urls}")

if __name__ == "__main__":
    cyclopts.run(cmd)
```

I expect both of these to work:

```bash
# Use as flag without arguments (should use defaults)
> python example.py --urls
urls: None

# Use with multiple arguments
> python example.py --urls url1 url2 url3
urls: ['url1', 'url2', 'url3']
```

Instead I get:

```bash
> python example.py --urls
╭─ Error ──────────────────────────────────────────────────╮
│ Command "cmd" parameter "--urls" requires an argument.    │
╰──────────────────────────────────────────────────────────╯
```

The second case (with arguments) works fine, but using it as a flag without arguments fails.

**Expected Behavior:**
When `consume_multiple=True` is used, the parameter should be able to:
1. Work as a flag without arguments (returning `None` or empty list)
2. Accept multiple arguments when provided

**Use Case:**
I think this pattern is common for CLI tools where a flag can enable a feature with defaults, or accept custom values when provided.

Is there a way to achieve this behavior with Cyclopts that I'm missing? Or a good workaround? Thank you!

BrianPugh:
Can you try your code on the `v4-develop` branch? I tweaked your example to:

```python
from typing import Annotated, Optional

from pydantic import BaseModel, Field

import cyclopts


@cyclopts.Parameter(name="*")
class Config(BaseModel):
    urls: Annotated[
        list[str] | None,
        Field(default=None, description="Optional list of URLs"),
        cyclopts.Parameter(
            consume_multiple=True,
        ),
    ]


def cmd(config: Optional[Config] = None):
    if config is None:
        config = Config()
    print(f"urls: {config.urls}")


if __name__ == "__main__":
    cyclopts.run(cmd)
```

And I **think** it works as expected. One thing that I did notice is that the following gets interpreted incorrectly:
```console
$ python issue-572.py --empty-urls
[...]

PR #578, opened 2025-10-06
Fix empty pydantic list

BrianPugh:
Addresses #572. @ilana-n please give this a try; i'm confident it will solve the issues you discovered!


ilana-n:
works great!!!! Thank you so much, this is very helpful.

Issue #598, opened 2025-10-13
Question about using a JSON object with cyclopts.config.Json


david-aimsun:
Hi,

Using the meta function, I can load a configuration file like this:

`app.config = cyclopts.config.Json( config )`

Here, config is a path to a JSON file. What if config is already a JSON object? I need to support configuration files that are read from the network.

Thanks,

David

BrianPugh:
this currently isn't supported, but let me see what I can do. I'm assuming by JSON object, you just mean data in a dictionary (e.g. similar to output of `json.load/json.loads`)

david-aimsun:
Yes, `json.loads` is in fact what I´m using.

Thanks!

PR #599, opened 2025-10-13
Add `cyclopts.config.Dict`. Add new keyword argument `source` to config objects.

BrianPugh:
Addresses #598.

@david-aimsun please give this a try!

### DIFF
--- a/tests/config/test_source_parameter.py
+++ b/tests/config/test_source_parameter.py
@@ -1,0 +1,181 @@
+"""Test that all config classes support the source parameter."""
+
+import json
+import os
+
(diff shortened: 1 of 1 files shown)


david-aimsun:
@BrianPugh It works ! Thanks a lot! This is what I did:

```python
@app.meta.default
def meta(
	config: Path = None,
):
	if config:
		app.config = cyclopts.config.Json( config )
	else:
		config_data = asyncio.get_event_loop().run_until_complete(remote_config())
		app.config = Dict( config_data )

	app(tokens)
```

Issue #605, opened 2025-10-15
[beta] Bash positional literal completion not working


contagon:
Hey, very excited about tab completions finally coming to cyclopts! I've been eager to switch to cyclopts from typer, but have been waiting for this final feature to do so.

I've hit a few completion edge cases that don't seem to be working properly (to be expected, it's in beta!). One of which is `Literal` positional argument completion doesn't seem to be working in bash (but it does in zsh).

Example:
```python
from typing import Literal
from cyclopts import App

app = App("cytest")
app.register_install_completion_command()  # type: ignore

@app.command
def command(param: Literal["foo", "bar", "baz"], /):
    """Simple command.

    Args:
        param (Literal["foo", "bar", "baz"]): literal param
    """
    print(f"Called {param}!")
```

This generates the following bash completion which doesn't mention foor, bar, or baz:
```
# Bash completion for cytest
# Generated by Cyclopts

_cytest() {
  local cur prev

  cur="${COMP_WORDS[COMP_CWORD]}"
  prev="${COMP_WORDS[COMP_CWORD-1]}"

  # Build list of options that take values (to skip their arguments)
  local options_with_values=''

  # Detect command path by collecting non-option words
  local -a cmd_path=()
  local i skip_next=0
  for ((i=1; i<COMP_CWORD; i++)); do
    local word="${COMP_WORDS[i]}"
    if [[ $skip_next -eq 1 ]]; then
      skip_next=0
      continue
    fi
    if [[ $word =~ ^- ]]; then
      # Check if this option takes a value
      if [[ " $options_with_values " =~ " $word " ]]; then
        skip_next=1
      fi
    else
      # Non-option word is a command
      cmd_path+=("$word")
    fi
  done

  # Determine command level and generate completions
  case "${#cmd_path[@]}" in
    0)
      if [[ ${cur} == -* ]]; then
        COMPREPLY=( $(compgen -W '--help -h --version --install-completion' -- "${cur}") )
      else
        COMPREPLY=( $(compgen -W 'command' -- "${cur}") )
      fi
      ;;
    1)
      case "${cmd_path[@]}" in
        "command")
          if [[ ${cur} == -* ]]; then
            COMPREPLY=( $(compgen -W '--help -h --version' -- "${cur}") )
          else
            COMPREPLY=()
          fi
          ;;
        *)
          ;;
      esac
      ;;
    *)
      ;;
  esac
}

complete -F _cytest cytest
```

[...]

PR #606, opened 2025-10-15
Bugfix: Fix bash choice-completion and positional FILES completion.

BrianPugh:
Addresses #605, @contagon can you give this a try?


contagon:
Awesome, this seems to fix the positional completion using bash! Works as expected for me now.

PR #613, opened 2025-10-17
Fix zsh positional-only choices

BrianPugh:
@contagon can you try this? Shell completion is so difficult to develop for (many environments, painful debug cycle, etc etc), so there's bound to be quite a bit of bugs. Thank you so much for the testing/error reporting!

### DIFF
--- a/cyclopts/completion/_base.py
+++ b/cyclopts/completion/_base.py
@@ -115,7 +115,16 @@ def get_completion_action(type_hint: Any) -> CompletionAction:
                     return action
         return CompletionAction.NONE
 
-    target_type = get_origin(type_hint) or type_hint
+    origin = get_origin(type_hint)
+
+    # For variadic positionals, unwrap tuple[T, ...] to get element type
+    if origin is tuple:
+        args = get_args(type_hint)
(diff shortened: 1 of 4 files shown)


contagon:
Happy to help! Thanks for all the quick fies. I can only imagine that's it's quite difficult with so many different shells/combinations possibles.

Alright, definite improvement, but it did introduce some new edges cases unfortunately. All tested in zsh,
- Completions now work when there is no docstrings,
- zsh positional list completion now works!
- unfortunately, it now accepts multiple arguments even when it is not a list, ie this
```python
@app.command
def command(param: Literal["foo", "bar", "baz"], /):
    print(f"Called {param}!")
```
will give me completions for `myapp command foo bar` which is invalid.
- keyword list arguments only give a single completion, ie this
```python
@app.command
def command(param: list[Literal["foo", "bar", "baz"]]):
    print(f"Called {param}!")
```
will complete for `myapp command --param foo` but not for an additional `--param` (not sure if it was this way before or not)

Let me know if you want

BrianPugh:
I'll try and fix that in a follow-up PR.

Issue #617, opened 2025-10-21
Mkdocs documentation generator


relativityhd:
With the release of v4 support for a Sphinx documentation generator was added.
Following #180 and especially the [comment from @danielgafni](https://github.com/BrianPugh/cyclopts/issues/180#issuecomment-3423751629) I want to open this issue to ask for mkdocs support.
Maybe if the foundations of such a generator are already here with Sphinx, this may also be possible in form of community contribution?

BrianPugh:
It shouldn't be too hard now that we have v4; I'll make a PR soon for this, but will certainly need testers before merging since I don't have mkdocs experience (but i've been intending to give it a try at some point).

relativityhd:
Sure, I can help to test it! Just ping me then, I should see it.

danielgafni:
Same! 

I've adopted Cyclopts for a new project I'm working on. Will be happy to test mkdocs.

danielgafni:
Overall it's working well for me! Here are some issues I've observed:
 
- TOC lacks nesting for subcommands

<img width="495" height="834" alt="Image" src="https://github.com/user-attachments/assets/eec4b12f-4880-4d6e-8eeb-74b38749a11d" />

- For some reason commands are truncated to a single word

<img width="585" height="1297" alt="Image" src="https://github.com/user-attachments/assets/05e0ef7f-9573-4d6c-a75a-7babaeea60ff" />
 
- Again, it would be nice to display env variables as well

Live url: https://anam-org.github.io/metaxy/main/reference/cli/

BrianPugh:
Released in v4.4.0! There will probably be a lot more tweaks that will need to be made, but let's make a new and separate issue for each one that comes up. It makes it much easier to track!

danielgafni:
Thanks for your hard work Brian!

Issue #477, opened 2025-06-24
Don't remove `Example`/`Examples` section from docstring


BrianPugh:
>We use the [docstring_parser project](https://github.com/rr-/docstring_parser) for parsing docstrings. It looks like Examples: gets [interpreted as a section](https://github.com/rr-/docstring_parser/blob/4951137875e79b438d52a18ac971ec0c28ef269c/docstring_parser/google.py#L48). Perhaps it would be best for Cyclopts to re-inject unused sections back into the main docstring body? Of these sections, I think only "Example/Examples" could be meaningful to the end-user?
> 
> I'll work on this next 

 _Originally posted by @BrianPugh in [#467](https://github.com/BrianPugh/cyclopts/issues/467#issuecomment-3001089052)_

BrianPugh:
ugh, this isn't going to be very clean; there's not a convenient way of specifying sections to parse/not-parse. I think it would require a PR/small refactor of upstream `docstring_parser`.

domdomegg:
MRE:

```python
#!/usr/bin/env python3

import cyclopts

app = cyclopts.App(
    name="test",
    help="Test CLI to understand cyclopts docstring handling",
)


@app.command
def command():
    """This is another test command.

    This command has a longer description in the second paragraph.
    We want to see how cyclopts handles multiple paragraphs.

    And here's a third paragraph for good measure.

    Examples:

        test command2 --option1 value1
        test command2 --option2 value2

    Notes:
        This is a notes section.
    """
    print("Command executed")

app()
```

Run with: `python script.py command --help`

domdomegg:
A workaround: rather than `Examples:` or `Notes:`, use something with a space e.g. `Usage examples:` or `Extra notes:`.

BrianPugh:
unfortunately, i don't think there's a good way to do this with docstring_parser, and I can't think of a good programmatic workaround. I think this is definitely a valid complaint, but i just don't think that anything can be done for now.

joelostblom:
Just chiming in to say that I think this would be a helpful feature. Thanks for picking up working on it again as per https://github.com/rr-/docstring_parser/issues/105#issuecomment-3835560277! How are you thinking that the Examples would be displayed in the help message, would it be an additional box/panel under the parameters?

BrianPugh:
[...]

PR #621, opened 2025-10-23
mkdocs plugin

BrianPugh:
Addresses #617 

@relativityhd @danielgafni testing and feedback would be appreciated!

### DIFF
--- a/tests/test_docs_snapshots.py
+++ b/tests/test_docs_snapshots.py
@@ -9,6 +9,9 @@ from pathlib import Path
 
 import pytest
 
+# Skip all snapshot tests on Windows due to platform-specific output differences
+pytestmark = pytest.mark.skipif(sys.platform == "win32", reason="Snapshot tests not supported on Windows")
+
 # Path to the complex-demo application
 COMPLEX_DEMO_DIR = Path(__file__).parent / "apps" / "complex-demo"
 


relativityhd:
Seems to work! I have tested it with our rather complex CLI which has a meta app, dataclasses and super nested commands.

I would suggest you get in contact with squidfunk to add cyclopts in the plugins schema file: https://squidfunk.github.io/mkdocs-material/schema/plugins.json so that yaml language server dont say that cyclopts is not a valid plugin

I've found one thing which is odd: Some of my commands have two "Options" headers where one is also present in the toc:

image

image

Same thing happens with arguments.

image

Always the second "Option" or "Argument" is without a ":" and is instead marked as a header.

This seems to only occur on commands which are part of double-nested apps. E.g. in my screenshots you can see that I have an app inside an app inside an app -> 3 subcommands

relativityhd:
Ah, and the filtering commands dont seem to work as I would expect:

```md

    :commands: training
    :exclude-commands: training.create-dataset
```

Will show all training commands, including the create-dataset command and subcommands.
On the other side:

```md
    :commands: training.create-dataset
/* no exclude */
```

Will hide the create-dataset sub-commands but show all other training commands.

BrianPugh:
thanks for the testing! I'll try and address these tomorrow.

danielgafni:
I will be testing it with my CLI tomorrow.

BrianPugh:
@relativityhd @danielgafni can you guys give this another round of testing? thanks!

danielgafni:
Definitely an improvement for env vars and the TOC entries! 

image

I think the last issue to address is the Markdown TOC which is still weird 
image

BrianPugh:
I'll get on it! Thanks for testing!

danielgafni:
Actually sorry, I was still running the old version. And confused positional arguments with env docs :man_facepalming: 

TOC is fine, only the missing env vars is an issue for me.

cyclopts/mkdocs_ext.py
```diff
@@ -0,0 +1,274 @@
...
+
+            This event is called after the page's markdown is loaded from file
+            but before it's converted to HTML.
+            """
+            if "::: cyclopts" not in markdown:
+                return markdown
+
+            return process_cyclopts_directives(markdown, self.config)
+
+else:
```
danielgafni: If you're looking for some comments: I think this runtime logic is not really needed. 
You should just never run the code that expects optional dependencies unless explicitly requested by the user. 
This removes a bunch of flaky error handling (e.g. other packages being the cause of `ImportError` you are trying to handle), duck typing, and so on. 

I prefer to do this the following way: 
- write your optional modules with normal Python imports and code
- for type annotations, move the optional types under `if TYPE_CHECKING` blocks
- for runtime usage, use lazy imports (inside functions and classes that actually use the optional deps) 
- make sure this module is not called *unless the user actually wants to use it*, e.g. never add it to parent `__init__.py` files, or have a plugin system that knows when to call it, and so on

cyclopts/mkdocs_ext.py
```diff
@@ -0,0 +1,274 @@
...
+
+try:
+    from mkdocs.config import base
+    from mkdocs.config import config_options as c
+    from mkdocs.exceptions import PluginError
+    from mkdocs.plugins import BasePlugin, get_plugin_logger
+
+    logger = get_plugin_logger(__name__)
+    MKDOCS_AVAILABLE = True
+except ImportError:
```
danielgafni: This entire logic can likely be dropped (see the other comment)

danielgafni:
Markdown TOC is actually fine now, sorry for the confusion.   

image

cyclopts/mkdocs_ext.py
BrianPugh: yup! totally agree; making this change now.

BrianPugh:
Thanks! I'm going to do a bit of a refactor (including your suggestions). I'm saving the env-var stuff for a separate PR because this is already quite a large PR, and it'll make it easier to review that functionality.

Once i have these changes in, i think we should have one more round of testing then merge. Then i'll work on the env var stuff!

danielgafni:
Understood! Thanks for your efforts Brian

relativityhd:
My previous reported bugs seem to be correct now.

But I found now several other problems, I tried to show them in this screenshot:
image

The mkdocs code for this is basically

```md
## Create Dataset Commands

::: cyclopts
    :module: darts.cli:app
    :commands: training.create_dataset
    :generate-toc: false
    :flatten-commands: true
    :heading-level: 3
```

Where training.create_dataset is a further subapp with multiple commands inside.

I know that we have a super nested CLI, it is awesome that it is possible to have such nesting, and it makes our lives so much easier!
I understand that this makes debugging of this PR super hard. So if you like, I can create a minimal reproducible version of our CLI if you need one. Or also help writing Code for it.

Either way: Thanks for your efforts Brian!

BrianPugh:
@danielgafni @relativityhd I'm kind of a noob at mkdocs, but i'm getting the vibe that the sphinx-like plugin syntax isn't idiomatic? Is it more common/prefered to use yaml syntax within the directive block? Because if so, i can easily switch to that (rather than emulating sphinx syntax).

relativityhd:
I think you can use [MkDocsStrings-Python](https://mkdocstrings.github.io/python/usage/) as an example, it is quite commonly used in the Python community as far as I can tell. Stuff in the directive block looks e.g. like this:

```
::: package.module.class
    options:
      do_something: false
```

But at the end, I think as long as you document how to do stuff, it doesn't matter a lot. I didn't notice the difference in declaration-syntax until you asked me about it. :D

danielgafni:
While it doesn't matter technically, it's better to conform to the current widespread style, IMO.

BrianPugh:
Yaml it is! I'll update this.

BrianPugh:
@danielgafni can you give this another round of testing?

@relativityhd same for you. Is there a branch of `darts` that I can play with to manually test some document generation? Thanks!

relativityhd:
Hi Brian, I've pushed my local changes to this branch:
https://github.com/awi-response/darts-nextgen/tree/feature/cycloptsv4

You would need to sync stuff with `uv sync --extra cpu --extra training --all-groups` to ensure all dependencies are loaded correctly.

Currently, we're still doing a lot of imports in the functions themselves to avoid long load times when calling --help on the CLI. I know that the new lazy-loading feature should help with that, but unfortunately this only works for nested subcommands, at least on another test I did. ("Lazy commands are resolved/imported in these situations: [...] Help Generation - When displaying help that includes the command") Maybe I will open another issue after having a closer look at this problem.

BrianPugh:
thanks for posting! I'll play around with doc-generation in the darts-nextgen repo.

As for the lazy-loading stuff, yes lets handle that separately in another issue/pr later!

BrianPugh:
@relativityhd I think we're getting closer; i'll continue to fiddle around with it, but I tweaked it to definitely be more in-line with what you would want.

danielgafni:
Hey Brian, I noticed these dataclass options are all rendered as both position arguments (`UPPERCASE`) and options (`--lower-case`), is this expected? 

In fact they are all optional. 

image

BrianPugh:
probably intentional! you probably want to mark your dataclass `kw_only=True` (or use [`KW_ONLY`](https://docs.python.org/3/library/dataclasses.html#dataclasses.KW_ONLY))

danielgafni:
Any idea if we could use [mkdocstrings](https://mkdocstrings.github.io/python/) from Cyclopts docstrings or parameters help?

## Better treatment for newlines in "long" help docstrings 

I've attempted to do something like this: 

```py
@app.command()
def my_command():
    """My short description. 

    My long description. 

    Example:
    
        $ my-command ...
    """
``` 

But the example got stripped from the generated markdown since the second newline ends the long description. 

Any idea if this behavior could be changed for markdown help format? 

## Document parameter types

I have this `dict[str, str]` argument but it's impossible to tell whether it accepts a dot-notation dict from the docs. I have to document this myself in the parameter help. 

image

Same issue stands with other types. Seems more like a general cyclopts problem since the CLI help is missing this info as well? Am I missing some configuration option here? 

##  Enable relative links to other docs 

Ideally I want to be able to link to my main docs from the CLI docs. This is typically done via relative links: 

```
[learn more here](../learn/concept)
```

However, this would break the urls displayed in CLI `--hepl`.

It sounds like this could be solved by allowing the user to specify a base URL pointing to the actual docs website and prefix the relative links with this base URL? Example: 

```py
app = cyclopts.App("app", docs_site_url="https://my.docs")

@app.command()
def my_command():
    """Learn more in [my docs](./learn/my-command)"""
```

```shell
$ app my-command --help
...
Learn more in [my docs](https://my.docs/learn/my-command)
```

danielgafni:
Hey @BrianPugh , could you please make a dev release from this branch?

BrianPugh:
sure! just a moment; I'm finally a bit more available after Thanksgiving madness, so I'll refocus on this PR!

BrianPugh:
this is now available as [v4.4.0a1](https://pypi.org/project/cyclopts/4.4.0a1/)! I'll have to review your above message soon and see what kind of improvements i can make.

danielgafni:
Thanks! 

I didn't realize until now this PR wasn't based on the `5.0.0` version of Cyclopts.

BrianPugh:
@danielgafni sorry for the slower iterations this time around, but here's what I'll be working on this morning:

> Any idea if we could use [mkdocstrings](https://mkdocstrings.github.io/python/) from Cyclopts docstrings or parameters help?

Not really (unless I'm completely misunderstanding); in order to work with everything else in Cyclopts, we have to be using our parsing system.

---

> But the example got stripped from the generated markdown since the second newline ends the long description.

I also share this frustration; this is because `docstring_parser` parses the section and we would somehow have to "unparse" it. We explore it a bit in #477. We also explore it in the docstring_parser repo a bit [here](https://github.com/rr-/docstring_parser/issues/105); but I honestly don't really have the will-power to deal with it. We could sort of hack-in support by find/replacing these section titles prior to docstring-parser, and then undo the find/replace afterwards; just seems quite fragile.

TLDR I agree!

---

>  it's impossible to tell whether it accepts a dot-notation dict from the docs

hmmm, yes this definitely does seem like a valid issue. What would be a good programmatic output? I agree it is a bit cumbersome to have to manually present an example. What if instead it instead displayed `-t.<KEY> --tags.<KEY>`? I'll tackle this in a separate, independent PR since it's a fairly independent issue.

---

> Ideally I want to be able to link to my main docs from the CLI docs.

I'll try and implement this feature now. It might be difficult, but I think it would be a good value add.
---

# My TODO list for this PR

Given all of the above, I think this PR is in a "good enough" state for a merge; we can continue to refine the idea in subsequent releases. However, I'll try and sneak in the docs-linking feature. What do you think @danielgafni @relativityhd ?

danielgafni:
> Not really (unless I'm completely misunderstanding); in order to work with everything else in Cyclopts, we have to be using our parsing system.

I mean this was an introduction to the rest of my post :) 

Being able to write mkdocstrings-compatible markdown is all we'd need, but the beeline issue makes it quite challenging. 

In general, I just want a way to attach mkdocs markdown to CLI commands in order for it to be rendered in the docs automatically. It doesn't necessarily need to be present in the Cli help (probably better if it's not since it can be large and confusing). 

Ideally I'd have CLI docs attached to CLI commands definitions so information stays local and automatically injected into the website.

BrianPugh:
ah! I'll investigate a little more; I was interpreting the original request as sort of overhauling the complete docstring parsing for some mkdocstrings solution. I'll keep you updated!

relativityhd:
image

I would't say it is working correctly with this PR, I just found another mis-representation of the markdown with the TOC. I also found some missing commands when playing around with the settings... Maybe my App is just too complicated, nested and full of edge-cases haha 😅 

However, mkdocs is now actually deprecated: https://squidfunk.github.io/mkdocs-material/blog/2025/11/05/zensical/ and the team behind mkdocs-material is building a new documentation framework.
That's said, I think investigating more time into this PR would be a waste of your time and skill. It works to a satisfying extent, it is usable and definitely better than nothing. So I want to thank you again for your great work! From my side, it would be fine to merge it.

I will open another Issues once Zensical becomes more mature and 3rd-party plugins (modules: https://zensical.org/about/roadmap/#module-system) are possible, because I think Zensical will become the next standard for technical documentation.

BrianPugh:
ah geez. Well, I guess time to start extracting the good collateral improvements from this PR into separate ones.

danielgafni:
Zensical is aiming to be a drop-in replacement, the markdown syntax is supposed to be compatible, right? 
I think 99% of this work is very much still relevant!

Besides, mkdocs itself is still likely to be used for many years ahead, and Zensical is currently very early. It didn't work that well in my project, and lots of plugins won't work with it right now.

BrianPugh:
alright @danielgafni @relativityhd, I just did another pass at this. Any more feedback is much appreciated! I figure that as @danielgafni mentioned, any mkdocs-successor is going to share a ton of code with what we implemented here. In-fact, the mkdocs-specific code is only like 250 lines, which is basically nothing.

In the spirit of timeboxing this, I want to reallly try and merge/release this by tuesday. It doesn't have to be perfect, but hopefully nothing breaking for subsequent enhancements.

In the darts-nextgen project, you'll probably want to leverage the newly added skip-preamble option in a few of your sections. Basically it skips the description/usage of the directly listed single command (in the example below, that would be `training.create_dataset`. This makes the resulting documentation look a lil better.

```yaml
## Create Dataset Commands

::: cyclopts
    module: darts.cli:app
    commands: [training.create_dataset]
    generate-toc: false
    flatten-commands: true
    heading-level: 3
    skip-preamble: true       # <----- this is new!
```

relativityhd:
Hi Brian, just tested it and it now works as I would expect, at least for my use-cases. Thank you! :D

Issue #623, opened 2025-10-23
Unable to add an epilog to the help output


isoschiz:
Another small missing feature that I have stumbled across while migrating from Typer to cyclopts is that there isn’t a way to print an epilog on the end of the help output.

https://typer.tiangolo.com/tutorial/commands/help/#epilog

(I generally use this to print version information about my tool on the end of the help output)

I did look at custom formatters (https://cyclopts.readthedocs.io/en/v4-develop/help_customization.html#creating-custom-formatters), but it doesn’t look like there’s a way to identify when the last panel is being written so it isn’t possible to tack on an extra line of content.

isoschiz:
I especially love that you have used the correct spelling rather than the (primarily American-English) misspelling.

BrianPugh:
closing just so i can keep track; this will be in the v4.1.0 release

PR #628, opened 2025-10-24
`App.help_epilogue` initial commit.

BrianPugh:
Addresses #623. @isoschiz what do you think of this PR?


isoschiz:
This looks great! Even better than Typer as it’s inherited by subcommands, whereas in my Typer app I had to repeat myself on every `app.add_typer()` call!

PR #635, opened 2025-10-24
Override ArgumentCollection __getitem__ to support looking up by parameter name

isoschiz:
In writing group validators, I was finding the fact that `ArgumentCollection.get()` returns `Optional[Argument]` to be a little cumbersome with type checking on. This PR overrides `__getitem__` as well, to provide a non-Optional function signature too.

### DIFF
--- a/cyclopts/argument/_collection.py
+++ b/cyclopts/argument/_collection.py
@@ -56,9 +56,7 @@ class ArgumentCollection(list[Argument]):
         if isinstance(term, (SupportsIndex, slice)):
             return super().__getitem__(term)
 
-        if (item := self.get(term)) is None:
-            raise KeyError(f"No such Argument: {term}")
-        return item
+        return self.get(term)
 
     @overload
     def get(


BrianPugh:
I think this is a welcomed addition! Can you review the changes I made? Thanks!

isoschiz:
Thanks to the code coverage warning about missed lines, I looked more carefully and I realised that `ArgumentCollection.get` already raises exceptions if the item can’t found! I think the actual issue is that the type annotation for that method isn’t correct.

In particular, it either returns an `Argument`, or it returns the value of `default`, which is specified as `Any`. So the true return type is `Argument | Any`. However, I’m not sure this is useful, since as a caller you would have to immediately check the type to determine which it was, and since you already had your default value to pass it in, you’re not gaining much here over just handling the exceptions and setting your default yourself (the caller probably means for the default to be the value of the argument, but if it is a true `Argument`, they’ll need to retrieve `.value` off it, which they won’t for their own default).

I don’t know what you’d prefer to do. My choice would probably be to change the default argument to `default: Argument | None = None`, and keep the return type of that function the same, but remove the raising of both IndexError and KeyError within it (i.e. just `return default` in the except path). Alternatively, you could change the type of default to just `Argument`, and then change the return type of `ArgumentCollection.get` to simply `Argument`, then change line 57 to just `return self.get(term)`. In both cases, callers will need to construct an `Argument` object to pass as a default for this function to be truly useful.

BrianPugh:
see if my latest commit fixes the type-hinting issues you've been running into. Regardless, I think it's good to implement the `__call__` method. I always intended to, just never got around to it!

isoschiz:
Thanks - I'll try to test this but might not be till Monday now. 

In any case, line 57 can still change to a direct return since .get() can't return None.

BrianPugh:
updated! just let me know when you get to it!

isoschiz:
Confirming that the type overloads on .get() resolve the mypy type complaint! Thanks for the speedy fix!

Issue #647, opened 2025-10-26
dataclass bare argument is silently ignored


isoschiz:
(doing some work with dataclasses so a couple of bugs incoming)

If I have a simple dataclass with a single field like this:

```
from typing import Annotated
from dataclasses import dataclass
import cyclopts

app = cyclopts.App()

@dataclass
class Foo:
    name: str

@app.default
def cmd(*, foo: Annotated[Foo, cyclopts.Parameter()] = Foo(name="foo")) -> None:
    print(foo)
```

Then if I pass a flag of `--foo.name`, I get the behaviour I expect. However, if I pass a bare `--foo` flag, then the value of it is silently ignored and we fall back to the default value:

```
$ uv run exp --foo.name hello
Foo(name='hello')
$ uv run exp --foo hello
Foo(name='foo')
```

Note that if the `foo` argument does not specify a default, then we correctly get an error when saying that `--foo.name` is a required argument.

Using `accepts_keys=False` in the annotation correctly supports `--foo` and produces an error for `--foo.name`, and similarly presents an error stating that `--foo` is required if the default value is removed.

BrianPugh:
that's a pretty serious bug; investigating now! Thanks again for finding and reporting all these bugs; I hope it doesn't discourage you from using Cyclopts 🙈 . With each bug report the library gets stronger!

BrianPugh:
Fixed in #651

PR #651, opened 2025-10-27
Allow positional elements of kw-only dataclass.

BrianPugh:
Addresses #647

### DIFF
--- a/tests/test_bind_dataclasses.py
+++ b/tests/test_bind_dataclasses.py
@@ -472,3 +472,39 @@ def test_dataclass_field_metadata_help(app, console):
 
     assert "Metadata help overrides docstring." in actual
     assert "This docstring is ignored." not in actual
+
+
+def test_bind_dataclass_direct_parent_match_issue_647(app, assert_parse_args):
+    """Test that --foo bar baz correctly converts to Foo(name="bar", age=25).
+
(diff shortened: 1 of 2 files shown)


tests/test_bind_dataclasses.py
```diff
@@ -472,3 +472,39 @@ def main(config: Config):
...
+    def cmd(*, foo: Annotated[Foo, Parameter()] = Foo(name="default", age=0)):  # noqa: B008
+        return foo
+
+    # This should work: directly provide values after parent name (positional-style)
+    assert_parse_args(cmd, "--foo bar 25", foo=Foo(name="bar", age=25))
+
+    # This should also still work: explicit nested keys
+    assert_parse_args(cmd, "--foo.name baz --foo.age 30", foo=Foo(name="baz", age=30))
+
+    # Mix of parent positional and nested key should work
```
isoschiz, line 503: This example is the same as the first one above, but this comment implies it should be a different test?

BrianPugh: whoops! good catch; this test case become stale as I was implementing; will remove it.

BrianPugh: removed in [cd5baa7](https://github.com/BrianPugh/cyclopts/commit/cd5baa790b3a3d8108e0e63994c22fe7cc9552b8)

Issue #648, opened 2025-10-26
dataclass(kw_only=True) with accepts_keys=False is not supported


isoschiz:
If I specify an argument using a dataclass which uses `kw_only=True` (which I often do as it makes the generated constructor more explicit), and use `accepts_keys=False` on that argument, as follows:

```
from typing import Annotated
from dataclasses import dataclass
import cyclopts

app = cyclopts.App(version_flags=())

@dataclass(kw_only=True)
class Foo:
    name: str

@app.default
def cmd(*, foo: Annotated[Foo, cyclopts.Parameter(accepts_keys=False)]) -> None:
    print(foo)
```

Then I get an error and can't construct the dataclass from the command line:

```
$ uv run exp --foo hello
Invalid value for --foo: Foo.__init__() takes 1 positional argument but 2 were given  
```

cyclopts needs to be able to pass the arguments as keywords, which would work for both the kw_only and regular flavours of the dataclass.

isoschiz:
As an aside: I tried using a custom converter to convert the token sequence into keyword arguments - but it doesn't look like the custom converter gets called in this scenario. I don't know whether that's intentional or not, but there was no error message or warning about the fact my converter would just be ignored.

BrianPugh:
fixed in v4.1.0

PR #658, opened 2025-10-28
Bugfix: kw-only dataclasses instantiation

BrianPugh:
Addresses #648.

@isoschiz can you give this a try?

### DIFF
--- a/cyclopts/_convert.py
+++ b/cyclopts/_convert.py
@@ -12,6 +12,7 @@ from typing import (
     TYPE_CHECKING,
     Any,
     Literal,
+    TypeVar,
     Union,
     get_args,
     get_origin,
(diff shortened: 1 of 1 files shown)


cyclopts/_convert.py
```diff
@@ -382,6 +381,115 @@ def _create_json_decode_error_message(
     return f"Invalid JSON for {type_.__name__}:\n    {snippet}\n    {' ' * marker_pos}^ {error.msg}{hint}"
 
 
+def instantiate_from_dict(type_: type, data: dict[str, Any]) -> Any:
```
isoschiz: You could use generics to improve the type signature here:

```
def instantiate_from_dict[T](type_: type[T], data[str, Any]) -> T:
    ...
```

PR #682, opened 2025-11-05
Avoid walruses in docs

thejcannon:
Since `.command(...)` re-returns the first argument, that means you can do `app_name = parent_app.command(App(...))` instead of `parent_app.command(app_name := App(...))` which avoids the walrus operator (i'm not a hater, I just try to avoid it when not strictly necessary. I do know there are some haters though, so let's give 'em less to hate)

### DIFF
--- a/docs/source/commands.rst
+++ b/docs/source/commands.rst
@@ -158,14 +158,14 @@ Subcommands inherit configuration from their parent apps.
    )
 
    # Child app inherits parent's settings
-   root_app.command(child_app := App(name="child"))
+   child_app = root_app.command(App(name="child"))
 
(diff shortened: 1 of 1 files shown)


BrianPugh:
This is great! It's obvious now that you presented it, but it hadn't occurred to me before to use the return value of `app.command`. I like this way a lot better! I'll do a quick skim of the docs and see if there's any remaining walrus references. Thanks!

thejcannon:
Glad you like it (and it works).

If there was any doubt, your typing of `command` is correct and type checkers seem to be just fine with this as well. So not an issue type-wise 👍

PR #683, opened 2025-11-06
Fix nested-meta-app command resolution.

### LINKED ISSUE #680: App commands are parsed out at wrong level with multiple meta apps
example.py:
```python
from typing import Annotated

app = App()


@app.meta.meta.default
def meta_meta_default(
):
    print("Meta meta default")
    command, bound, _ignored = app.meta.parse_args(tokens)
    command(*bound.args, **bound.kwargs)


@app.meta.default
def meta_default(
    meta_param: str = "default",
):
    print("Meta default")
    app(tokens)


@app.default
def default():
    print("Default")


@app.command
def command():
    print("command")


if __name__ == "__main__":
    app.meta.meta()

```

Everything works as expected with no command:

```
$ python example.py
meta meta default
meta default
default
```

When using the command, `app.meta` is skipped:

```
$ python example.py command
meta meta default
command
```

But, it works again if any `app.meta` parameters are used:

```
$ python example.py --meta-param value command
meta meta default
meta default
command
```

For better typing, I would write `meta_meta_default` as:

```python
@app.meta.meta.default
def meta_meta_default(
):
    print("meta meta default")
    _command, bound, _ignored = app.meta.parse_args(tokens)
    meta_default(*bound.args, **bound.kwargs)
```

Now, `command` doesn't even run:

```
$ python test.py command
meta meta default
meta default
default

$ python example.py --meta-param  value command
meta meta default
meta default
command
```

This also means that `--install-completion` can't be accessed if created using `register_install_completion_command`.

BrianPugh:
Should fix #680.  @JuneStepp can you please test? Thanks for reporting!

### DIFF
--- a/cyclopts/core.py
+++ b/cyclopts/core.py
@@ -226,7 +226,7 @@ def _combined_meta_command_mapping(
                 command_mapping[cmd_name] = subapp[cmd_name]
 
     if recurse_meta and app._meta:
-        command_mapping.update(_combined_meta_command_mapping(app._meta))
+        command_mapping.update(_combined_meta_command_mapping(app._meta, recurse_parent_meta=False))
     if recurse_parent_meta and app._meta_parent:
         meta_parent_commands = _combined_meta_command_mapping(app._meta_parent, recurse_meta=False)
         command_mapping.update(meta_parent_commands)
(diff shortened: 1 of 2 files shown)


JuneStepp:
This works. Thanks!

fyi, I've also noticed that `--help` errors with `meta.meta` commands. It does work with `meta`  commands but doesn't show `meta.meta` parameters. Non-meta commands also don't show `meta.meta` parameters. Only `meta` ones. I can open an issue for that later if needed.

BrianPugh:
I'll investigate/fix that in a separate PR now, thanks!

Issue #685, opened 2025-11-11
Short isoformats not supported for `datetime`


jonaslb:
E.g. `2025-11-11T13Z` (no minute, second or fractional second). This is supported by `datetime.fromisoformat`.

Maybe this would be easier if I could specify that the Pydantic validation rules should be used for arguments instead?

BrianPugh:
Fixed in #687.

> Maybe this would be easier if I could specify that the Pydantic validation rules should be used for arguments instead?

Can you elaborate on this? Conversion is defered to pydantic if you are within a pydantic BaseModel.

jonaslb:
> Can you elaborate on this?

I actually expected the pydantic validation logic for datetimes to be used, instead of cyclopt's own. Maybe my relative unfamiliarity with the library is showing (picked it up recently due to unhappiness around typer), but I also was hoping to use `pydantic.Field` annotations for function arguments? Maybe if I had done that, cyclopts would have parsed the datetime with pydantic?

BrianPugh:
Fixed in v4.2.3

PR #687, opened 2025-11-11
use `datetime.fromisoformat`

BrianPugh:
Addresses #685


jonaslb:
Suggestion: Instead of using a hardcoded time format in the fallback case, you could try to do "string repair", e.g. `datetime.fromisoformat(s.strip().replace(" ", "T"))`. Otherwise, with the currently hardcoded format, you're missing allowance for with/without fractional seconds and timezones.

Alternatively, you could drop Python 3.10 but that might be too wild (although it's what i'd do) 😆

Issue #704, opened 2025-12-04
Feat: Ignore arguments as examplef or dependency injection


jochumdev:
Hi!

Thanks for this great peace of software, I'm having a lot of fun with all the features cyclopts provides!

I'm using <https://python-dependency-injector.ets-labs.org/wiring.html> for a project, it uses an "@inject" decorator to overwrite arg defaults.

Sadly cyclopts also wants to handle them which results in wired behavior.

This little patch allows me to ignore fields with an underscore prefix, I don't think this is backwards compatible that's why I didn't provide a PR.

```diff
diff --git a/cyclopts/argument/_collection.py b/cyclopts/argument/_collection.py
index 87abf08..00b5dfe 100644
--- a/cyclopts/argument/_collection.py
+++ b/cyclopts/argument/_collection.py
@@ -435,6 +435,9 @@ class ArgumentCollection(list[Argument]):
         docstring_lookup = extract_docstring_help(func) if parse_docstring else {}
         positional_index = 0
         for field_info in signature_parameters(func).values():
+            if not field_info.required and field_info.name[0] == "_":
+                continue
+
             if parse_docstring:
                 subkey_docstring_lookup = {
                     k[1:]: v for k, v in docstring_lookup.items() if k[0] == field_info.name and len(k) > 1
```

What do you think about this?

Kind regards,
René

jochumdev:
@BrianPugh dang you are fast!

BrianPugh:
thanks for opening this issue! I'll investigate it locally. Regardless, I can imagine ignoring "private" function parameters could be a common/generically useful feature. We'll have to have this skipping be disabled-by-default for now, but we can make it enabled-by-default in v5.

jochumdev:
Can provide you a copy of my project (it will be open source) so you can play with that.

BrianPugh:
Alright, after investigating/doing-a-lil-bit-of-research, here are my thoughts:

1. I think it would be great that if parameters that begin with an underscore default to `Annotated[..., Parameter(parse=False)]` in Cyclopts (i.e. we skip them).
2. This would be a breaking change, so we would need to wait for v5.
3. Drawing from other popular libraries for inspiration:
   * [attrs](https://www.attrs.org/en/stable/init.html#private-attributes-and-aliases) always strips the leading underscore. If the user really wants the leading, they can reintroduce it via an `alias`.
[...]

PR #705, opened 2025-12-05
Allow `Parameter.parse` to be a regex string (only parse those that match the regex).

BrianPugh:
Addresses #704. @jochumdev can you play with this? Basically for your use-case (and what should be the default for cyclopts v5) what you want is:

```python
# Only parse parameters that do NOT start with an underscore "_"
app = App(default_parameter=Parameter(parse="^(?!_)"))
```

## TODO
- [x] Documentation
- [x] Compile the regex for performance

### DIFF
--- a/cyclopts/argument/_argument.py
+++ b/cyclopts/argument/_argument.py
@@ -186,6 +186,14 @@ class Argument:
                 )
 
         if not self.parse:
+            # Validate that non-parsed parameters are keyword-only or have defaults
+            is_keyword_only = self.field_info.kind is self.field_info.KEYWORD_ONLY
+            has_default = self.field_info.default is not self.field_info.empty
+            if not (is_keyword_only or has_default):
+                raise ValueError(
+                    f"Non-parsed parameter '{self.field_info.name}' must be a KEYWORD_ONLY function parameter "
+                    "or have a default value."
+                )
             return
 
         if self.parameter.accepts_keys is False:
--- a/docs/source/default_parameter.rst
+++ b/docs/source/default_parameter.rst
@@ -74,3 +74,97 @@ When resolving what the :class:`.Parameter` values for an individual function pa
 5. **Lowest Priority:** (2-4) recursively of the parenting app call-chain.
 
 Any of Parameter's fields can be set to `None` to revert back to the true-original Cyclopts default.
+
+.. _Skipping Private Parameters:
+
+---------------------------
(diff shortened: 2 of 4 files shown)


jochumdev:
> Addresses https://github.com/BrianPugh/cyclopts/issues/704. @jochumdev can you play with this? Basically for your use-case (and what should be the default for cyclopts v5) what you want is:

Works, nice work!

<details><summary>main.py</summary>

```py
	from pathlib import Path
	from typing import Annotated

	from rich.pretty import pprint

	from dependency_injector import containers, providers
	from dependency_injector.wiring import (
	    Provide,
	    inject,
	    register_loader_containers,
	)


	class Container(containers.DeclarativeContainer):
	    config_path = providers.Dependency(Path)

	    config = providers.Configuration(
	        strict=True,
	    )


	app = App()


	@app.default
	@inject
	def test(_config: dict[str, any] = Provide(Container.config)):
	    pprint(_config, expand_all=True)


	@app.meta.default
	async def meta(
	    config: Path = Path("./config.json"),
	) -> None:
	    """
	    A Command-Line Interface.

	    Args:
	      config: Configuration file
	    """

	    ct = Container(config_path=config)
	    ct.config.from_json(config)

	    if future := ct.init_resources():
	        await future

	    ct.check_dependencies()
	    ct.wire(modules=[__name__], warn_unresolved=True)

	    # This needs to be before ANY dynamic python imports.
	    register_loader_containers(ct)

	    # app.command("mypkg.commands.subapp1:app", name="subapp1")

	    rc = await app.run_async(tokens=tokens)

	    if future := ct.shutdown_resources():
	        await future

	    return rc


	if __name__ == "__main__":
	    app.meta()
```
</details> 

<details><summary>config.json</summary>

```json
	{
	  "foo": "bar"
	}
```

</details> 

<details><summary>pyproject.toml</summary>

```toml
	[project]
	name = "cylopts-skip-parse"
	version = "0.1.0"
	description = "Add your description here"
	requires-python = ">=3.13"
	dependencies = [
	  "cyclopts",
	  "dependency-injector>=4.48.3",
	]

	[tool.uv.sources]
	cyclopts = { git = "https://github.com/BrianPugh/cyclopts.git", branch = "skip-parse" }
```

</details> 

run with:
```sh
uv run main.py
```

BrianPugh:
this will be a part of v4.4.0 (targeting a release in <1week)

Issue #712, opened 2025-12-17
How to pass None explicitly for optional parameter with default value?


Alerion:
I have `write_concurrency: int | None = 2` parameter and can't get how to pass `None` explicitly. I tried `--write_concurrency=null` and `--write_concurrency=None`, but I get error `Invalid value for "--write_concurrency": unable to convert "null" into int.`

BrianPugh:
You will have to configure the [`Parameter.negative_none` flag](https://cyclopts.readthedocs.io/en/stable/api.html#cyclopts.Parameter.negative_none) or the [`Parameter.negative` flag](https://cyclopts.readthedocs.io/en/stable/api.html#cyclopts.Parameter.negative).

I just realized that the documentation for `Parameter.negative` doesn't mention this; i will update now.

Alerion:
It would be helpful to pass [required](https://github.com/BrianPugh/cyclopts/blob/main/cyclopts/argument/_argument.py#L939) to `converter` in `Parameter`. Then have something like this:
```
def parse_type_or_none[T](type_: type[T], tokens: Sequence[Token], required: bool) -> T | None:
    token = tokens[0]
    if token.value == "none" and not required:
        return None
    return type_(token.value)

App(
     default_parameter=Parameter(
         converter=parse_type_or_none
    )
)
```
Now I can have a custom convertor for every optional parameter, but I like that cyclopts does not force me to wrap everything into `Annotated` to keep code clean.

But `negative_none=("none-",)` also works, so it is not a blocker. It is just confusing that all values can be passed after `=`, but only for `None` you should remember that it is a prefix.

BrianPugh:
so you probably wouldn't want that conversion functions for your entire app; it would work for simple bare types (str, int, etc), but I don't think it would work for more complex types and unions.

> It is just confusing that all values can be passed after =, but only for None you should remember that it is a prefix.

This is a little bit of an oversight; `None` support was added fairly recently. However, generally, it feels "wrong" to expose `None` to CLI consumers. Both in your example (and in general within the context of Cyclopts CLI Applications), `None` is a sentinel value. In your case, I can imagine it would could be better to use a type like `int | Literal["default"]` and then handle the `"default"` string in your command body. I'm sure your full example is more complicated, but I guess my main point is that there might be better ways to go about this instead of using `None`.

Alerion:
> It is just confusing that all values can be passed after =, but only for None you should remember that it is a prefix.

[...]

PR #713, opened 2025-12-17
document optional/none behavior in Parameter.negative

BrianPugh:
Improves documentation for #712. @Alerion does this seem helpful?

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -999,6 +999,17 @@ API
 
       Set to an empty list or string to disable the creation of negative flags.
 
+      .. note::
+
+         Setting :attr:`negative` to **any** value (including an empty list/string) overrides
+         **all** automatic prefix generation from :attr:`negative_bool`, :attr:`negative_iterable`,
+         and :attr:`negative_none`. For example, ``negative=()`` disables negative flags for
+         the entire parameter, even if ``negative_none="none-"`` is also set.
+
+         To selectively disable only one type of negative prefix while keeping others functional,
+         set the specific attribute to an empty string instead. For example, use ``negative_bool=""``
+         to disable boolean negatives while allowing :attr:`negative_none` to still generate flags.
+
       Example usage:
 
       .. code-block:: python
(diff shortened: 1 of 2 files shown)


Alerion:
Thank you for your quick response. 

It is useful, but it does not work with:
```
negative=(),
negative_none="none-",
```

Some context. I have many different python functions(task functions). And I add them automatically to cyclopts app. So all options are function parameters. So I try to force to use `false`, `true` and `none` explicitly, so it looks like similar to a function call. Because usually you just see a task function and want to copy-paste parameters into console. And using prefixes is confusing, because you need to map python parameters into console options with prefixes.

Also, I don't want to explicitly add `negative` to every task function, because it reduces readability.

I will check a source code and try to find solution.

BrianPugh:
Ah! you've identified another gap in the documentation; `Parameter.negative` overrides all of the other automatically generated negative values (e.g. `negative_bool`, `negative_iterable`, `negative_none`). I'll add documentation about this.

> So I try to force to use false, true and none explicitly

Unfortunately, i think Cyclopts won't support none-flags for a `bool | None`; it'll only generate the true/negative boolean flags. In the union case, we only generate the none-flag if there are no bool/iterable flag candidates.

> And using prefixes is confusing, because you need to map python parameters into console options with prefixes.

I both agree and disagree; like it or not, it's the idiomatic way of handling flags for CLI applications. Cyclopts does allow you to specify true/false for boolean flags if-and-only-if you use an equal sign (not a space) between the flag and value. E.g. `--my-flag=false`. It has to be this way because a flag doesn't consume the next space-delimited token.

> Also, I don't want to explicitly add negative to every task function, because it reduces readability.

You can define `App(default_parameter=Parameter(negative=...))` to define the defaults you want at a larger scope.

If you can copy/paste a minimal script of what you are trying to achieve, I can help! It'll make the discussion a little more concrete!

BrianPugh:
updated PR with notes about how `Parameter.negative` overrides the other automatic negative-flag generation logic.

BrianPugh:
closing this in favor of #717

PR #717, opened 2025-12-22
Allow none/null strings to be directly parsed into `None`. Support Union of different-token-length types.

BrianPugh:
Addresses #712.

- Add support for parsing `"none"` and `"null"` strings (case-insensitive) to `None` for optional types.
- Union parsing now includes `None`/`NoneType` (previously skipped).
- Support Union of different-token-length types. Became a higher priority now that `none` is a common single-token option.

This PR is based on v5-develop because this could catch a minority of users by surprise.

@Alerion what do you think of this?


Alerion:
I think this should completely solve my issues. Thx

BrianPugh:
Just a heads up: I'm a bit slow on merging this because I want to improve it's ability to generalize with multi-token type hints, like we would want `None | tuple[int, int]` to be parsed as `None` if the next token is `none`/`null`.

PR #732, opened 2026-01-12
Bugfix: Unknown parameters in config fails for user classes

### LINKED ISSUE #731: Unknown parameters in config fails for user classes
The following script fails with error `Parameter [np][timeout] specified multiple times.`.

```python

import cyclopts

app = cyclopts.App(
    config=cyclopts.config.Dict(
        {"p": {"timeout": 3}, "np": {"timeout": 4}},
        allow_unknown=True,
        use_commands_as_keys=False,
    )
)


@dataclass
class ConnectionParameter:
    timeout: float


@app.default
def connect(*, p: ConnectionParameter):
    pass


app()
```

Setting `allow_unknown=False` does not change the outcome and it also fails with the same error when replacing the `dataclass` with `pydantic.BaseModel`.

I was excepting that the extra `np` key in the config to be ignored and the command to run successfully.

With the config dict `{"p": {"timeout": 3}}`, the `connect` function is successfully called with the expected parameter `p`. But, more surprisingly, with the config dict `{"p": 3, "np": {"timeout": 4}}` the function is also called with the same parameter and no error message.

BrianPugh:
Fixes #731; @AntoninRousset can you give this a try? thanks!

### DIFF
--- a/cyclopts/argument/_collection.py
+++ b/cyclopts/argument/_collection.py
@@ -596,6 +596,38 @@ def _meta_arguments(apps: Sequence["App"]) -> ArgumentCollection:
     return argument_collection
 
 
+def _is_valid_option_key(option_key: str, arguments: "ArgumentCollection") -> bool:
+    """Check if option_key corresponds to a valid root argument.
+
+    When processing nested config keys like {"p": {"timeout": 3}}, the fallback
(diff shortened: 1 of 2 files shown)


AntoninRousset:
I have tested and it fixes #731, as well as its side issue.

Thank you very much!

PR #745, opened 2026-02-13
List correct values when unsupported value passed to Enum param

### LINKED ISSUE #744: Include possible values for Enum members in error message
When using `Enum`, cyclopts helpfully shows the possible member values in the the help text, but they do not show up in the error message when an incorrect value is passed via the CLI (as they do for `Literal`).

Take this app as an example:

```python
from cyclopts import App
from enum import Enum
from typing import Literal

app = App()


class NumberStrings(Enum):
    one = "one"
    two = "two"
    three = "three"


@app.default
def default(
    number_string: NumberStrings | None = None,
    number: Literal["1", "2", "3"] | None = None,
):
    print(number_string, number)


app()
```

The `--help` command shows the available choices for both the string literal and the `Enum`:

Image

A typo when specifying the `--number` parameter, helpfully lists the possible correct values that can be passed:

Image

However, this is not currently the case for `Enum`, where the error message just mentions that "ones" is not part of the `NumberStrings` class:

Image

This information is not as helpful as for the string literal and it is difficult for the user of the CLI to know how to fix the error. I think it would be more helpful if the Enum member names were printed similarly to how they are included in the help text:

Image

I played around a bit and the following changes to the code would change to the improved error message for Enums:

```diff
if get_origin(self.target_type) is Literal:
    choices = "{" + ", ".join(repr(x) for x in get_args(self.target_type)) + "}"
    target_type_name = f"one of {choices}"
+ elif isinstance(self.target_type, type) and issubclass(self.target_type, Enum):
+     choices = "{" + ", ".join(repr(x) for x in + self.target_type._member_names_) + "}"
+    target_type_name = f"one of {choices}"
else:
    target_type_name = get_hint_name(self.target_type)
```

If you think this seems like a desired change and a reasonable approach, I can explore the change a bit deeper and open a PR.

joelostblom:
This PR shows the available values of a CLI parameter when an incorrect value is passed. Closes #744

### DIFF
--- a/tests/test_coercion.py
+++ b/tests/test_coercion.py
@@ -14,6 +14,7 @@ import pytest
 
 from cyclopts import CoercionError, Token
 from cyclopts._convert import convert, token_count
+from cyclopts.utils import default_name_transform
 
 
 def _assert_tuple(expected, actual):
@@ -171,8 +172,8 @@ def test_coerce_enum():
 
 def test_coerce_enum_invalid_choice():
     class GroupedConstants(Enum):
-        foo = auto()
-        bar = auto()
+        FOO = auto()
+        BAR = auto()
 
     assert_convert_coercion_error(
         GroupedConstants,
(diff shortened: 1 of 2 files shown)


tests/test_coercion.py
```diff
@@ -169,6 +169,54 @@ class SoftwareEnvironment(Enum):
         convert(SoftwareEnvironment, ["invalid-choice"])
 
 
+def test_coerce_enum_invalid_choice():
+    class GroupedConstants(Enum):
+        foo = auto()
+        bar = auto()
```
joelostblom: I replicated the tests for the string literal error message, let me know if you think something else is needed. I also just duplicated the Enum class, but can create a constant if you prefer.

BrianPugh:
Thanks for opening this PR! I think i'm realizing that this feature might actually be slightly more complicated, as we need to apply `name_transform` to the options. E.g. if we changed one of the enums to be:

```python
    class GroupedConstants(Enum):
        FOO = auto()
        BAR = auto()
```

then the listed values would be capitalized:

```
Invalid value for "MOCKED_ARGUMENT_NAME": unable to convert "invalid-choice" into one of {'FOO', 'BAR'}.
```

I'll try and take a look at this closer this weekend to figure out what to do.

BrianPugh:
@joelostblom I made some changes that I think address the `name_transform` issue, can you also review/play around with it? Thanks!

joelostblom:
Thank you @BrianPugh! I played around a bit and it seems to be working as expected for me :+1:

PR #746, opened 2026-02-17
Reference correct callable in entrypoints example

joelostblom:
There is no callable `main` in the cyclopts example code. Rather we need to reference the `app` callable to be launch via the created entrypoint. This might be clear for someone familiar with how to create entrypoints in Python, but for someone who is newer to these concepts, I think this change can be helpful.


BrianPugh:
I think that's fair, thank you!

Issue #750, opened 2026-02-20
Strict option for long option parsing (--long-option=thing, --long-option thing)


sspencer-arine:
Ahoy.   This is a bit opinionated.  Submitting a request to alter how long option parsing works if possible so that for any option that requires a value that `=` is required.   Not a huge priority - just wanting to support POSIX as well as the GNU long option standard.

https://sourceware.org/glibc/manual/2.30/html_node/Argument-Syntax.html#:~:text=GNU%20adds%20long%20options%20to,option%20names%20in%20the%20shell.

BrianPugh:
I can take a peek to see how easy/hard this would be to implement in the current code-base. I personally always prefer using `=`. I'll take a look sometime within the next few days and let you know my findings!

BrianPugh:
released in v4.6.0

PR #751, opened 2026-02-21
New Feature: `Parameter.requires_equals`

BrianPugh:
Addresses #750. @sspencer-arine can you give this a try? Thanks!

### DIFF
--- a/cyclopts/argument/_argument.py
+++ b/cyclopts/argument/_argument.py
@@ -186,6 +186,13 @@ class Argument:
                     f"Use 'Annotated[int, Parameter(count=True)]' for counting flags."
                 )
 
+        if self.parameter.requires_equals and self.parameter.consume_multiple:
+            raise ValueError(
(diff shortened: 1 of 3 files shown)


sspencer-arine:
@BrianPugh this is great!

I like how intentional it is, and someday I might lobby for having this set from the application, requiring explicit handling if this needs to be false in any situations. I think this is a wonderful patch for what is most likely a corner case right now and we can see where it grows if at all.

BrianPugh:
I'll release this soon. You can always enable this at the app-level via:

```python
app = App(
    default_parameter=Parameter(requires_equals=True),
)
```

and it should "just work" as expected.

sspencer-arine:
Ah very clever..   I'm learning a lot about cyclopts right now and TIL this ;)  Thanks @BrianPugh

sspencer-arine:
Already a big hit here:

```
Parameter "--output" requires a value assigned with "=". Use "--output=VALUE". 
```

Issue #754, opened 2026-02-23
Show: A way for command chaining without delimiters


Tremeschin:
<sup>Note: This would fit better in a show and tell discussion~ (disabled)</sup>

Hi,  I'm testing the feasibility of cyclopts over typer as I can more easily define commands from attrs or pydantic classes and their docstrings, but command chaining is a must in my projects, preferably keeping compatibility (no delimiter syntax).

I came up with a short `@app.meta.default` method to split at known command names:

```python
def chainer(app: App) -> callable:
    def meta(*tokens: Annotated[str, Parameter(show=False, allow_leading_hyphen=True)]):

        # Find known commands indices or default empty split
        splits = [n for (n, token) in enumerate(tokens) if token in app] or [0]

        # Split and call all groups of tokens
        for start, end in itertools.pairwise(splits + [len(tokens)]):
            _result = app(tokens[start:end])

    return meta
```

```python
if __name__ == "__main__":
    app = App()
    app.command(foo)
    app.command(bar)
    app.meta.default(chainer(app))
    app.meta(sys.argv[1:])
```

Surely there are limitations and side effects I'm not yet aware, requires non-conflicting arguments down the chains with meta app. Just thought sharing for maybe adding to docs or others to find this 🙂

BrianPugh:
I think this is a neat use-case, i'll add something like this to the "cookbook" section of the docs. I like to put patterns like this there (rather than integrated as first class features) if the configuration/limitations/tradeoffs a user might want to make is quite large/complex. Because at that point, having them directly copy/paste the code is more powerful (so long as it's not too much code!).

PR #756, opened 2026-02-23
And 'without delimiter' example to docs.

BrianPugh:
Addresses #754.

@Tremeschin what do you think about this?

### DIFF
--- a/docs/source/command_chaining.rst
+++ b/docs/source/command_chaining.rst
@@ -58,15 +58,15 @@ If your command names and argument values never collide, you can split tokens at
    app = App()
 
 
-   def split_commands(app: App, tokens: tuple[str, ...]) -> list[list[str]]:
+   def split_commands(app, tokens):
        """Split tokens into groups, starting a new group at each known command name."""
-       groups: list[list[str]] = []
+       groups = []
        for token in tokens:
            if token in app:
                groups.append([])
            if groups:
                groups[-1].append(token)
-       return groups
+       return groups or [[]]
 
 
    @app.command


Tremeschin:
Looks good, but I think there's an edge case I stumbled on my own previously

When arguments are empty, `app([])` shows help and usage, but `app.meta([])` doesn't

When no groups can be made, nothing is run. I "solved" that by making an empty group via `splits + [0]` index in the related issue, in this PR that would roughly translate to `return groups or [[]]`

BrianPugh:
fixed! I'll merge this soon unless you have any objections/feedback!

Tremeschin:
No objections, only but a feedback that it comes down to preference to either have the auxiliary `split_commands` or everything in `main(*tokens)` - I personally prefer the later, to reduce footprint

Yours solution doesn't depend on `itertools`, is easier for beginners (no `list + list` syntax or list comprehension, pairwise range grouping), all fine by me!

Tremeschin:
Ah, I just noticed something, this ignores/allows unknown arguments before a group (in mine too):

`app.meta(["unknown", "items", "foo", "123", "bar", "--flag"])`

Should we force a group to start at the first index or raise exceptions if not known command?

Tremeschin:
Solution: Simply start with an empty group! 🙂 

```python
def split_commands(app, tokens):
    groups = [[]]
    for token in tokens:
        if token in app:
            groups.append([])
        groups[-1].append(token)
    return groups
```

Tremeschin:
Sorry minor spam, looks good inlined too and is much simpler

```python
@app.meta.default
def main(*tokens: Annotated[str, Parameter(show=False, allow_leading_hyphen=True)]):
    group = []
    for token in tokens:
        if token in app:
            app(group)
            group.clear()
        group.append(token)
    app(group)
```

The last call handles both the last group that doesn't match a new command and empty runs

PR #757, opened 2026-02-27
avoid resolving lazy commands during --help

zzstoatzz:
## Context

We're using cyclopts' lazy loading for Prefect's CLI (~29 subcommands). While profiling startup time, we found that `format_usage()` resolves all lazy commands via `_registered_commands` → `__getitem__`, even though `groups_from_app()` avoids resolution with its `resolve_lazy` parameter (from #710).

As you [pointed out](https://github.com/BrianPugh/cyclopts/pull/757#issuecomment-3975227056), the initial fix had an unintended side effect: lazy commands disappeared from the help page entirely, since `groups_from_app(resolve_lazy=False)` skips unresolved commands and `format_usage` was no longer resolving them either.

## What this PR does

Two changes that work together:

1. **`format_usage`**: replaces the resolving visibility check with the simpler approach you suggested — checking whether any iterator name isn't a help/version flag.

2. **`CommandSpec` + `groups_from_app`**: adds optional `help` and `show` fields to `CommandSpec` so metadata can be provided at registration time (e.g. `app.command("mod:func", name="cmd", help="...")`). `groups_from_app` creates lightweight stub Apps for unresolved lazy commands, so they appear in help output with their pre-specified help text — no module import needed.

This way, lazy commands show up in `--help` (with descriptions if `help=` is provided at registration), without triggering resolution.

## Test plan

- `test_format_usage_does_not_resolve_lazy_commands` — regression test for the `format_usage` path
- `test_lazy_command_non_leaf_help_does_not_resolve` — the case you identified, now passing (lazy children listed without resolution)
- `test_groups_from_app_resolve_lazy` — stubs appear with pre-specified help text
- `test_groups_from_app_lazy_hidden` — `show=False` commands excluded
- Updated existing tests that assumed `--help` would resolve lazy commands
- Full test suite passes (2012 passed, 35 skipped)

### DIFF
--- a/cyclopts/help/help.py
+++ b/cyclopts/help/help.py
@@ -272,7 +272,9 @@ def format_usage(
     for command in command_chain:
         app = app[command]
 
-    if any(app[x].show for x in app._registered_commands):
+    # Check for non-help/version commands without resolving lazy CommandSpecs.
+    help_version_flags = {*app.help_flags, *app.version_flags}
+    if any(x not in help_version_flags for x in app):
         usage.append("COMMAND")
 
     if app.default_command:
--- a/cyclopts/command_spec.py
+++ b/cyclopts/command_spec.py
@@ -42,6 +42,8 @@ class CommandSpec:
     import_path: str
     name: str | tuple[str, ...] | None = None
     app_kwargs: dict[str, Any] = Factory(dict)
+    help: str | None = None
+    show: bool = True
 
     _resolved: "App | None" = field(init=False, default=None, repr=False)
 
--- a/cyclopts/core.py
+++ b/cyclopts/core.py
@@ -1313,7 +1313,16 @@ class App:
 
             # Create CommandSpec with the resolved name (first name if multiple)
             # The name will be used when wrapping functions in an App
-            spec = CommandSpec(import_path=obj, name=name[0] if name else None, app_kwargs=kwargs)
+            # Extract CommandSpec-specific fields from kwargs before storing
+            spec_help: str | None = kwargs.pop("help", None)  # type: ignore[assignment]
+            spec_show: bool = kwargs.pop("show", True)  # type: ignore[assignment]
+            spec = CommandSpec(
+                import_path=obj,
+                name=name[0] if name else None,
+                app_kwargs=kwargs,
+                help=spec_help,
+                show=spec_show,
+            )
 
             # Register the CommandSpec
             for n in name + alias:
--- a/cyclopts/group_extractors.py
+++ b/cyclopts/group_extractors.py
@@ -82,6 +82,20 @@ def groups_from_app(app: "App", resolve_lazy: bool = False) -> list[tuple[Group,
         cmd = app._get_item(name, recurse_meta=True)
         if isinstance(cmd, CommandSpec) and not cmd.is_resolved:
             if not resolve_lazy:
+                if not cmd.show:
+                    continue
+                # Create a lightweight stub App for help display without importing
+                cmd_id = id(cmd)
+                app_names.setdefault(cmd_id, []).append(name)
+                if cmd_id not in unique_apps:
+                    from cyclopts.core import App as _App
+
+                    unique_apps[cmd_id] = _App(
+                        name=name,
+                        help=cmd.help or "",
+                        version_flags=[],
+                        help_flags=[],
+                    )
                 continue
         subapp = app[name]
         app_id = id(subapp)
--- a/tests/test_group_extractors.py
+++ b/tests/test_group_extractors.py
@@ -96,7 +96,12 @@ def test_same_app_registered_multiple_times():
 
 
 def test_groups_from_app_resolve_lazy():
-    """Test that resolve_lazy parameter controls whether lazy commands are resolved."""
+    """Test that resolve_lazy parameter controls whether lazy commands are resolved.
+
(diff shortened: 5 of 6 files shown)


cyclopts/help/help.py
```diff
@@ -272,7 +272,29 @@ def format_usage(
     for command in command_chain:
         app = app[command]
 
-    if any(app[x].show for x in app._registered_commands):
```
BrianPugh, line 275: I think we can simplify all of this and get (nearly) the expected functionality with:

```python
# No need to do `or ()`; app.help_flags and app.version_flags are always tuples despite the type hint.
help_version_flags = set(app.help_flags + app.version_flags)
if any(x not in help_version_flags for x in app):
     usage.append("COMMAND")
```

This won't check if the command's `show=True`, but this would only impact the case where there are ONLY hidden subcommands for an app, which seems like an unrealistic edge-case. Additionally, this only impacts the help-visualization, so the stakes are a bit low. Also, this is arguably more correct behavior, but honestly it's around 50/50.

BrianPugh (commented):
Thanks for the PR! I added just a small comment. I haven't tested the end-to-end behavior, but i'll do that when you mark this as ready-for-review.

cyclopts/help/help.py
zzstoatzz: makes sense! thanks for explaining

zzstoatzz:
thanks for the quick review @BrianPugh - let me know if you'd like to see anything else here :)

BrianPugh:
I think this all looks good; i'll merge and do a patch release today. But, just to explicitly confirm, this fixed the lazy-loading for your CLI?

zzstoatzz:
> But, just to explicitly confirm, this fixed the lazy-loading for your CLI?

affirmative!

<details>
<summary>Benchmarks and import tracking</summary>

**Before this fix**, `format_usage()` resolves all lazy commands via `_registered_commands` → `__getitem__`:

```
$ uv run prefect --help
Traceback (most recent call last):
  ...
  File "cyclopts/help/help.py", line 275, in format_usage
    if any(app[x].show for x in app._registered_commands):
  File "cyclopts/core.py", line 866, in __getitem__
    return cmd.resolve(self)
  ...
```

Every lazy `CommandSpec` gets resolved (imported) just to check visibility.

**After this fix**, `--help` works without resolving any lazy commands:

```
$ hyperfine --warmup 2 --runs 10 'uv run prefect --help'
Benchmark 1: uv run prefect --help
  Time (mean ± σ):     724.6 ms ±  46.0 ms    [User: 624.6 ms, System: 90.4 ms]
  Range (min … max):   686.9 ms … 832.3 ms    10 runs
```

```
$ uv run python -c "
import sys
cli_imports = []
original_import = __builtins__.__import__
def tracking_import(name, *args, **kwargs):
    if name.startswith('prefect.cli.') and name not in ('prefect.cli._app', 'prefect.cli', 'prefect.cli._utilities') and name not in sys.modules:
        cli_imports.append(name)
    return original_import(name, *args, **kwargs)
__builtins__.__import__ = tracking_import
from prefect.cli._app import _app
_app.meta(['--help'], exit_on_error=False, result_action='return_value')
print('CLI modules imported during --help:', cli_imports or 'none')
"
CLI modules imported during --help: none
```

</details>

BrianPugh:
@zzstoatzz I'm playing around with this command a bit, and i think it has the unintended side effect of not listing lazy-loaded commands in the help-page (see `test_lazy_command_non_leaf_help_does_not_resolve`). I think we really need to unfortunately resolve commands to get the command's short-help-string (unless it's provided at the command-registration).

zzstoatzz:
oops! good catch, i moved this back to draft

i will go back and think about this a bit more, sorry for the runaround!

zzstoatzz:
hmm, what do you think about https://github.com/BrianPugh/cyclopts/pull/757/commits/2cf52f85a0927cd7b92c221fa02f447d0361df4d? im a bit dissatisfied w how big the diff got, but was just thinking through this

BrianPugh:
just letting you know i haven't forgotten about this, just been a bit tied up at the moment, I'll try to get to it soon!

zzstoatzz:
> just letting you know i haven't forgotten about this, just been a bit tied up at the moment, I'll try to get to it soon!

no worries! i can handle my needs in user-space for now without too much fuss :)

BrianPugh:
@zzstoatzz what do you think of [this](https://github.com/BrianPugh/cyclopts/tree/fix/format-usage-lazy-resolution)? For some reason it's not letting me push to your branch, and I gave up trying 😅 . I'm going to re-review my solution in the morning, but it's what i'm currently leaning towards.

zzstoatzz:
> @zzstoatzz what do you think of [this](https://github.com/BrianPugh/cyclopts/tree/fix/format-usage-lazy-resolution)? For some reason it's not letting me push to your branch, and I gave up trying 😅 . I'm going to re-review my solution in the morning, but it's what i'm currently leaning towards.

👁️ at a glance it looks clean! but i'll try depending on this tomorrow and see how it goes and let you know!

thanks for the update 🫶

zzstoatzz:
@BrianPugh works like a charm! do you want me to adopt that here or just close this and let u merge it otherwise?

BrianPugh:
I can handle it from here (probably will just open another PR). I didn't actually get a chance to re-review this morning, but I will tonight.

BrianPugh:
closing in favor of #761

Issue #89, opened 2024-01-25
[Feature Request] Completion?


Pack3tL0ss:
Curious if this project currently has completion (couldn't find it in the docs or code) or if there are aspirations to provide completion in the future??

BrianPugh:
I would love to add autocompletion, but have no experience in the area. I would certainly welcome any input, whether it be general recommendations, library suggestions, PRs, etc.

BrianPugh:
Just a heads up, I am working on this, hoping to have something testable in the next week or two.

BrianPugh:
shell completion is infinitely more complex than I was initially expecting 😄 . Still a WIP.

mattangus:
This is somewhat of a deal breaker for me so I'm keen to help out!

BrianPugh:
Currently I'm working on adding bash and zsh support; i'm taking another direction on it and going to be basing the code off of [shtab's implementation](https://github.com/iterative/shtab/blob/main/shtab/__init__.py), which I think is a good reference.

Originally I was pursuing the idea of adding functionality to generate an `argparse.ArgumentParser` object from a `cyclopts.App`. However, some Cyclopts concepts don't directly translate (such as positional-or-keyword arguments). The benefits of this approach is that other off-the-shelf tooling could be used. I'm still considering this approach, but it does have its issues.

BrianPugh:
I'm still thinking about this; I'm actually leaning towards the argparse solution (and maybe even also having a `to_click` method later) where I map postional-or-keyword arguments to be keyword-only to make them more compatible with those libraries/ecosystems.

alextremblay:
Hi @BrianPugh 

I've recently discovered cyclopts after becoming disillusioned with typer's slow (possibly stalled) development

I love everything you've done with it so far, and the only thing preventing me from adopting it in all my projects is the lack of shell completion. I have users leveraging shell completion in typer now, and taking that away from them is a hard no

I'm wondering if you'd be willing to publish your WIP on shell completion in another branch so that I or others could take a look at it and possibly help you finish it?

alextremblay:
nvm, I just realized you already **did** that! hahaha

looking at the code, i think i see what you were going for. I agree that some kind of parser transformer (converting and `App` to an `ArgumentParser` or a click `Group` would be better

BrianPugh:
Hey @alextremblay! I totally agree that shell completion should be the next important feature to implement

[...]

PR #758, opened 2026-03-04
better 'just-works' behavior for oh-my-zsh users.

BrianPugh:
My solution to the discussion [here](https://github.com/BrianPugh/cyclopts/issues/89#issuecomment-3977173517).

> I'm using oh-my-zsh and I had to move the completion script from the default install location ~/.zsh/completions to "$ZSH_CUSTOM"/completions. I also removed the call to compinit from my .zshrc. This was on my WSL2 installation. Everything works great now.

I want Cyclopts to "just work" for the majority of users, and that seems like changing the install locaiton for oh-my-zsh users. It's a small amount of code and not much of a maintenance burden. Shell completion is always tricky, and so it's nice if the user doesn't have to think about it too much.

@onyx-and-iris I was wondering if you could try/review this? Thanks!

### DIFF
--- a/cyclopts/completion/install.py
+++ b/cyclopts/completion/install.py
@@ -4,12 +4,40 @@ This module handles the installation of completion scripts to shell-specific
 locations and the updating of shell RC files to load completions.
 """
 
+import os
 import sys
 from collections.abc import Callable
 from pathlib import Path
 from typing import Annotated, Literal
 
 from cyclopts.parameter import Parameter
+
+
+def _detect_omz_completions_dir() -> Path | None:
+    """Detect oh-my-zsh custom completions directory.
+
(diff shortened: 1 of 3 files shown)


onyx-and-iris:
Yes I can, I'll get back to you later today on it.

onyx-and-iris:
I can confirm this is working in my testing:

```console
onyx@hostname ➜  ~ vban-cli --install-completion
✓ Completion script installed to /home/onyx/.oh-my-zsh/custom/completions/_vban-cli
✓ Detected oh-my-zsh: completions directory is already in $fpath.

Restart your shell or run: exec zsh
```

after restarting the shell my completions were working!

Thanks.

onyx-and-iris:
I've raised a separate issue on the topic of shell completions: https://github.com/BrianPugh/cyclopts/issues/759

Issue #762, opened 2026-03-09
Allow a way to reject empty list arguments when using `consume_multiple=True`


ShaiAvr:
I really like working with `cyclopts`. You have done a great job. However, in my recent project, I had a use case that doesn't seem possible to implement at the moment. I wanted to write a kind of solution checker that accepts a solution program to a specified problem and checks the program against test cases that are specified as files of input and expected output. A required parameter is, of course, the problem to solve, and by default, the program will scan the corresponding directory and test the solution against all the test cases it finds. I wanted to implement an option to select which test cases to run instead of all of them, so I decided to add a `--cases` option that allows the user to specify which test cases to run instead of running all of them:
```python
# /// script
# requires-python = ">=3.14"
# dependencies = [
#     "cyclopts>=4.8.0",
# ]
# ///

from typing import Annotated

from cyclopts import App, Parameter

app = App()


@app.default
def main(problem: str, /, *, cases: list[str] | None = None) -> None:
    print("Checking solution for problem:", problem)
    print("Test cases:", cases if cases is not None else "All cases")


if __name__ == "__main__":
    app()
```
By default, when `cyclopts` sees a list keyword argument, it creates an `--empty-*` option to allow the user to pass an empty list. In my use case, it doesn't make sense to pass an empty list of test cases, because if you don't have any test cases to check, then you don't need to run the checker in the first place. Luckily, it's easy to disable this feature:
```python
# /// script
# requires-python = ">=3.14"
# dependencies = [
#     "cyclopts>=4.8.0",
# ]
# ///

from typing import Annotated

from cyclopts import App, Parameter

app = App()


@app.default
def main(
    problem: str,
    /,
    *,
    cases: Annotated[
        list[str] | None, Parameter(negative_iterable="")
    ] = None,
) -> None:
    print("Checking solution for problem:", problem)
    print("Test cases:", cases if cases is not None else "All cases")


if __name__ == "__main__":
    app()
```
Now, by default, `cyclopts` parses a list option by specifying the keyword argument multiple times:
```console
$ check_solution demo_problem --cases simple_test --cases another_test
Checking solution for problem: demo_problem
Test cases: ['simple_test', 'another_test']
```
[...]

Issue #763, opened 2026-03-09
Parsing a positional list argument continues after an option


ShaiAvr:
While playing around with [positional list arguments](https://cyclopts.readthedocs.io/en/stable/rules.html#positional), I noticed a behaviour which I think is weird and could be a bug. To demonstrate, consider this demo program:
```python
# /// script
# requires-python = ">=3.14"
# dependencies = [
#     "cyclopts>=4.8.0",
# ]
# ///

from cyclopts import App

app = App()


@app.default
def main(foo: list[str] | None = None, /, *, bar: int = 0, baz: int = 0) -> None:
    print("Foo:", foo)
    print("Bar:", bar)
    print("Baz:", baz)


if __name__ == "__main__":
    app()
```
Now, suppose I start feeding the `foo` argument, then I supply the keyword arguments, and then I try to keep feeding the `foo` argument, which works just fine:
```console
$ my-program a b c --bar 8 --baz 10 d
Foo: ['a', 'b', 'c', 'd']
Bar: 8
Baz: 10
```
I think this command should result in an error since we finished parsing the values of `foo` when we encountered `--bar`, and anything that comes after that shouldn't belong to `foo` anymore, and in this example, this makes `d` an undefined argument. I certainly wouldn't expect the value "d" to be part of the `foo` list because I have already given the values for `foo` before the options. It would have made sense if I had given the options before supplying `foo`: `$ my-program --bar 8 --baz 10 a b c d`. However, I'd argue that even this syntax isn't very good because the rule of Python is that positional arguments must come before keyword arguments. If I were to call `main` in the script, I'd have to pass `foo` positionally before passing `bar` or `baz` by keyword. I think the command line syntax should reflect that.

I think passing keyword arguments before positional arguments on the command line looks wrong, so I'd personally reject `$ my-program --bar 8 --baz 10 a b c d`, and only allow `$ my-program a b c d --bar 8 --baz 10`. However, I suppose it's subjective, and there's no need to change the current behaviour. We can use a compromise like introducing a flag `force_positional_before_keyword`, which is `False` by default (and thus retaining the current behaviour), and when it's `True`, it will make `cyclopts` reject keyword arguments before positional arguments on the command line.

[...]

PR #764, opened 2026-03-10
Allow `consume_multiple` to be an `int` or a `tuple[int, int]`. 

BrianPugh:
This sets lower & upper bounds on number of elements to consume.

Addresses #762. @ShaiAvr what do you think of this solution? In your case, it would be `Parameter(consume_multiple=1)`, indicating that you need at least 1 element.


ShaiAvr:
@BrianPugh This looks like a great solution, which solves my use case and also others.

I also want to say that I just noticed another behaviour which I think should be configurable somehow. When I first read the docs of the `consume_multiple` parameter, I thought that setting it to `True` would prevent passing the option multiple times and instead force the user to pass the option once with the list of values. I just found out that even with `consume_multiple=True`, the user can still pass the option multiple times, and he's simply able to pass multiple values to the option each time. I assume this PR conserves this behaviour. I don't know if this behaviour is intentional, but if it is, then I think there should be a way to configure it.

My original intention was to reject passing the option multiple times and only allow passing it once with the list of all the values. However, it seems like `cyclopts` doesn't support limiting the number of times a list option can be passed.

BrianPugh:
> I don't know if this behaviour is intentional, but if it is, then I think there should be a way to configure it.

It's intentional, but I understand the desire to configure it. We can tackle it in another PR! I'll keep it in mind as I try to figure out a reasonable solution to #763. Could you make a github issue to track this? It doesn't need to be as thorough as your other issues. Thanks!

ShaiAvr:
@BrianPugh I created the issue #765.

Issue #770, opened 2026-03-13
Custom parser error formatter


tahv:
Hi,

Cyclopts uses `CycloptsPanel` to format its validation error messages. This panel has a certain look that can't be changed easily.

```text
$ my-script foo.bin
╭─ Error ───────────────────────────────────────────────────────╮
│ Invalid value "foo.bin" for "SRC". "foo.bin" does not exist.  │
╰───────────────────────────────────────────────────────────────╯
```

In the other parts of my app, for example, errors and warning messages have a different look:

- `"[bold red]error[/bold red]: my error message"`
- `"[bold yellow]warning[/bold yellow]: my warning message"`

Could you consider exposing a hook to format cyclopts validation errors ? In my case, the example above would look like this:

```text
$ my-script foo.bin
error: Invalid value "foo.bin" for "SRC". "foo.bin" does not exist.
```

From my search, `CycloptsPanel` is only used [once in the library](https://github.com/BrianPugh/cyclopts/blob/ead15046cc464beef16a63859ec41b917062cbcf/cyclopts/core.py#L1775).

## Example implementation

My idea is to add `App.message_formatter`, a callable that takes a message and returns a printable object. That callable would default to `CycloptsPanel` to maintain compatibility.

```python
from typing import Any, Protocol

class MessageFormatter(Protocol):
      def __call__(self, message: Any, title: str = "Error", style: str = "red") -> Any: ...

class App:
    message_formatter: MessageFormatter = field(default=CycloptsPanel, kw_only=True)
```

And call the formatter in cyclopts instead of using `CycloptsPanel` directly.

```python
class App:

   def parse_args(...):
      ...
         except CycloptsError as e:
             ...
             if print_error if print_error is not None else True:
                 e.console.print(self.message_formatter(e))
```

In my app, the callable would look something like this:

```python
def MyMessageFormatter(self, message: Any, title: str = "Error", style: str = "red") -> str:
   title = title.lower()
   if not bold in style:
      style = f"bold {style}"
   return f"[{style}]{title}[/{style}]: {message}"
```

If that's something you are interested in, I can possibly work on it and propose a PR.

Thanks.

BrianPugh:
added in v4.10.0!

PR #771, opened 2026-03-14
allow for custom CycloptsError formatters.

BrianPugh:
addresses #770. @tahv what do you think of this? Thanks!


tahv:
Looks great, thank you !

PR #787, opened 2026-04-09
[help] Add granular help for dict[str, BaseModel] params

wrongbad:
https://github.com/BrianPugh/cyclopts/issues/786

What do you think?

### DIFF
--- a/cyclopts/argument/_argument.py
+++ b/cyclopts/argument/_argument.py
@@ -431,6 +431,71 @@ class Argument:
             else self._match_name(term, transform=transform, delimiter=delimiter)
         )
 
+    def _normalize_trailing_keys(self, trailing: tuple[str, ...]) -> tuple[str, ...]:
+        """Map kebab-case segments back to their canonical Python field names.
+
+        Walks the type hint segment-by-segment:
+
+        * Dynamic ``dict`` keys pass through unchanged (advance to the value type).
+        * Segments addressing a structured type (pydantic / dataclass / attrs /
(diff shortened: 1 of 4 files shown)


BrianPugh:
thanks for the PR! I only had a chance to skim it right now, but I like that the diff is minimal! However, I think this would only work 1 level deep, right? I think we may need it to be recursive.

wrongbad:
Do you think the single-level depth is useful enough to merge a first PR, and revisit arbitrary recursion later?

BrianPugh:
I'll take a look at this tonight; if there's not a straightforward way of having the recursive support, we could merge this as-is as a partial fix.

wrongbad:
I took a stab at recursive support. See the updated test_help for how I interpreted that. The diff in help.py is still within reason. I didn't bother with lists, or immediate nesting (`dict[str, dict[str, MyClass]]`). Let me know if you want that too.

BrianPugh:
sounds good, thanks! sorry about the slow turn around time on this; i've been traveling work, i'm hoping to work on this this weekend.

BrianPugh:
I just pushed some changes to make this work more generically (not just with pydantic). Does it still work properly in your project?

wrongbad:
Yeah the tests cases are still good. I also just tested with my project, and I'm seeing all the help args and aliases I'd hope to see. Thanks for collaborating on this one!

BrianPugh:
awesome! i'll merge this in and prep for a release for the next day or two! Thanks for the PR!

Issue #802, opened 2026-05-11
Feature Request: Support string/classmethod references for `validator=` (matching `converter=` behavior)


gil-ch-palo:
## Description

`Parameter(converter=)` supports string references to classmethods, allowing self-contained type definitions where the conversion logic lives inside the class itself. However, `Parameter(validator=)` does not support this pattern, requiring validators to be defined as standalone functions outside the class.

It would be great if `validator=` supported the same string reference resolution that `converter=` already has, enabling self-contained validation logic via classmethods.

## Motivation

When using `@attrs.define` (or `@dataclass`) classes as CLI parameter groups, it's natural to keep validation logic inside the class. However, to get cyclopts' formatted error panel (the `╭─ Error ─╮` box), validation must go through `Parameter(validator=)`. Since `validator=` requires a callable reference at decoration time, and the class body isn't defined yet at that point, the validator function must be defined outside and before the class. This separates the validation logic from the data it validates.

An alternative like `__attrs_post_init__` keeps the logic inside the class, but the resulting `ValueError` bypasses cyclopts' error handling entirely and produces a raw traceback instead of the formatted error panel.

### Current workaround (validator as external function)

```python
import cyclopts as cyp
import attrs
from typing import Annotated


def _validate_color_mix(type_, value) -> None:
    if value.bold and value.italic:
        raise ValueError("Cannot use both --bold and --italic together.")


@cyp.Parameter(name="*", validator=_validate_color_mix)
@attrs.define
class TextStyle:
    color: Annotated[str, cyp.Parameter(alias="-c")]
    bold: Annotated[bool, cyp.Parameter()] = False
    italic: Annotated[bool, cyp.Parameter()] = False


app = cyp.App()


@app.default
def main(style: TextStyle):
    print(f"Color: {style.color}, Bold: {style.bold}, Italic: {style.italic}")


app()
```

### Desired (matching converter= pattern)

```python
import cyclopts as cyp
import attrs
from typing import Annotated


@cyp.Parameter(name="*", validator="validate")
@attrs.define
class TextStyle:
    color: Annotated[str, cyp.Parameter(alias="-c")]
    bold: Annotated[bool, cyp.Parameter()] = False
    italic: Annotated[bool, cyp.Parameter()] = False

    @classmethod
    def validate(cls, value) -> None:
        if value.bold and value.italic:
            raise ValueError("Cannot use both --bold and --italic together.")


app = cyp.App()


@app.default
def main(style: TextStyle):
[...]

PR #803, opened 2026-05-12
resolve string forward-referenced validators

BrianPugh:
addresses #802. @gil-ch-palo can you give this a try? thanks!


gil-ch-palo:
Tested it with our codebase, worked as expected! 🫡 

Thanks @BrianPugh for the quick implementation.

Issue #805, opened 2026-05-15
Error messages for Literal/Enum coercion failures unintuitive for inexperienced users


jwinpbe:
hi brian!

I'm designing a program for people to control remote devices with, and for the auth portion, I have `pyshock init`, which has an alias of "auth". I wired it up with an alias referencing this function:

```python
def init(
    provider: Annotated[
        Literal["pishock", "openshock"] | None,
            Parameter(...)
        # etc
```

So in the help, it appears like `init (auth)  Set up your account.`

trying to anticipate a user who is not good with computer, i ran `pyshock init auth` and got:

`Invalid value for "PROVIDER": unable to convert "auth" into one of {'pishock', 'openshock'}.`

Imagining someone who doesn't program at all seeing this, the first thing I think of is that they have no idea what they're 'converting', dont recognize 'PROVIDER' (an internal value), and don't know what to do.

I can use `Parameter(validator=_auth_validator)` and extract it into a small function, but given that `Literal` is supported as an Annotated parameter type, I was wondering if you would be ok with adding something like `invalid_choice_message` as an option you can pass through Parameter to change the error message.

The lying machine that lives in my graphics card told me that if we added:

- An `invalid_choice_message` or something else semantically clear after `help` in parameter.py (around line 250)
- Literal and Enum imports to _argument.py
- ~7 lines of logic in the exception block around line 825 to check for the new param and do str.replace() on {token} / {choices} if present, most of which is no-op if the new param is `None`

then I could pass

```python
Parameter(invalid_choice_message='Unknown provider "{token}". Please choose from one of: {choices}')
```

and have it come back as:

`Unknown provider "auth". Please choose from one of "pishock" / "openshock"`

The lying machine seems convinced that "all of the wiring is already there, and `Literal[]` has first class support, so it makes sense to add", so I was wondering if you'd be open to a PR on this or if you'd rather just have me use my own validator function.

thanks!

-- jwin

BrianPugh:
> Imagining someone who doesn't program at all seeing this, the first thing I think of is that they have no idea what they're 'converting', dont recognize 'PROVIDER' (an internal value), and don't know what to do.

I think this is good feedback. I think we have two, possibly mutual, decisions/solutions to make here:

[...]

PR #806, opened 2026-05-16
friendlier error message for enum/literal choices

BrianPugh:
## Summary

Two changes to runtime error formatting:

1. **Friendlier `CoercionError` for `Literal` / `Enum` types.** Leads with the offending value, drops the "unable to convert ... into one of {...}" jargon and Python-set notation, and adds a `Did you mean X?` suggestion via `difflib` (matching `UnknownCommandError` / `UnknownOptionError`).
2. **Dropped quotes around parameter / command / group names and "Did you mean" suggestions across every error class.** `PROVIDER`, `--my-flag`, and `--bar` are already visually delimited by uppercase / leading dashes; the quotes added noise without disambiguation. Quotes are kept around actual user-input values (where arbitrary characters are possible).

Addresses #805. @jwinpbe what do you think of this?

## Before / after

Given `Literal["pishock", "openshock"]` and `pyshock init auth`:

**Before**
```
Invalid value for "PROVIDER": unable to convert "auth" into one of {'pishock', 'openshock'}.
```

**After**
```
Invalid value "auth" for PROVIDER. Choose from: "pishock", "openshock".
```

**After (near-miss triggers suggestion)** — `pyshock init pisock`:
```
Invalid value "pisock" for PROVIDER. Choose from: "pishock", "openshock". Did you mean "pishock"?
```

| Before | After |
|---|---|
| `Parameter "--bar" requires an argument.` | `Parameter --bar requires an argument.` |
| `Command "foo" parameter "--bar" requires an argument.` | `Command foo parameter --bar requires an argument.` |
| `Unknown command "fuzz". Did you mean "fizz"?` | `Unknown command "fuzz". Did you mean "fizz"?` |
| `Unknown option: "--fooo". Did you mean "--foo"?` | `Unknown option: "--fooo". Did you mean --foo?` |
| `Invalid values for group "foo".` | `Invalid values for group foo.` |
| `Parameter "--name" requires a value assigned with "=".` | `Parameter --name requires a value assigned with `=`.` |

Non-Literal/Enum coercion errors (`int`, `Path`, etc.) keep `unable to convert ... into <type>` — that phrasing is still accurate for those types.

### DIFF
--- a/cyclopts/core.py
+++ b/cyclopts/core.py
@@ -2790,6 +2790,6 @@ def _log_framework_warning(framework: TestFramework) -> None:
         else:
             var_name = matched_app_variables[0]
 
-        message = f'Cyclopts application invoked without tokens under unit-test framework "{framework.value}". Did you mean "{var_name}([])"?'
+        message = f'Cyclopts application invoked without tokens under unit-test framework "{framework.value}". Did you mean {var_name}([])?'
         warnings.warn(UserWarning(message), stacklevel=3)
         break
--- a/cyclopts/core.py
+++ b/cyclopts/core.py
@@ -2790,6 +2790,6 @@ def _log_framework_warning(framework: TestFramework) -> None:
         else:
             var_name = matched_app_variables[0]
 
-        message = f'Cyclopts application invoked without tokens under unit-test framework "{framework.value}". Did you mean {var_name}([])?'
+        message = f'Cyclopts application invoked without tokens under unit-test framework "{framework.value}". Did you mean "{var_name}([])"?'
         warnings.warn(UserWarning(message), stacklevel=3)
         break
--- a/cyclopts/exceptions.py
+++ b/cyclopts/exceptions.py
@@ -154,14 +154,14 @@ class ValidationError(CycloptsError):
             except IndexError:
                 pass
             else:
-                provided_by = "" if not token.source or token.source == "cli" else f' provided by "{token.source}"'
+                provided_by = "" if not token.source or token.source == "cli" else f" provided by {token.source}"
                 name = token.keyword if token.keyword else self.argument.name.lstrip("-").upper()
-                message = f'Invalid value "{value}" for "{name}"{provided_by}.'
+                message = f'Invalid value "{value}" for {name}{provided_by}.'
         elif self.group:
             if self.group.name:
-                message = f'Invalid values for group "{self.group.name}".'
+                message = f"Invalid values for group {self.group.name}."
         elif self.command_chain:
-            message = f"Invalid values for command {self.command_chain[-1]!r}."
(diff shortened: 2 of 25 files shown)


jwinpbe:
its a lot better thank you!! wiring in difflib is great.

i like it as is. Here's some bike sheds that i worked thru because i'm underemployed:

<details>
<summary>i really like rich terminal style aesthetics</summary>

> 
> for `pyshock init pisock` would it be easy to do:

> `Invalid value "pisock" for PROVIDER. Choose from: "pishock", "openshock". Did you mean [bold]"pyshock init pishock"[/bold]?`

i'm thinking of people who would reach for a copy paste to fix the typo and would be so unfamiliar with a terminal they wouldn't hit the up arrow and would start re typing.

i have watched at least 5 people re-type command typos over screen share and had to coach them to up arrow until they learned. bless their heart.

> i grepped and only found `.upper()` here and in a help display. my style inclination is towards `PROVIDER` being [bold].capitalize()[/bold] in the above but it's not my baby.

then, upon further review, I realized we can't do the rich formatting as it sits, but if you added a check for `__rich__` in a passed `message` from a raising error in [panel.py](https://github.com/BrianPugh/cyclopts/blob/b0c66c452139de521f0bab4260eeda2af5ca15e9/cyclopts/panel.py#L37), you could pass it's `__rich__` through as the renderable instead of `__str__`, falling through to the current implementation if `__rich__` wasn't there. you could duplicate the `__str__` logic in `__rich__` (or helper method it) and add stuff like dimming / bold / italic / `text_color on color` for the printed errors. 

I'd be willing to do something like that if you'd read the PRs one class at a time. I'm certainly using cyclopts enough on my own to want to make it pretty.

My profession isn't SWE but I won't just get claude to do it.
</details>

thanks again!!!

BrianPugh:
> My profession isn't SWE but I won't just get claude to do it.

Hey now, professionals also just get Claude to do it 😂 . We just make sure to actually carefully review/think if the proposed solution is correct.


Here are my thoughts:

1. I'm going to pass on the copy-pastable did-you-mean. It seems nice in this toy example, but it could quickly get way too verbose and hit a bunch of edge-cases.
2. I agree that something like "Provider" could be a bit more natural in some use-cases over "PROVIDER", but all-caps generally has positional-meaning in the CLI world, and it doesn't look too out-of-place for people less familiar with CLIs.
3. I think sprinkling more `rich` in is always a good thing, **i'll open up a followup PR to handle that.**

> I'd be willing to do something like that if you'd read the PRs one class at a time. I'm certainly using cyclopts enough on my own to want to make it pretty.

Sure, as you see new things that you think you can tackle, i'm always very open to PRs! If unsure, you can always open up an issue first!

Issue #817, opened 2026-05-23
Add support for `**kwargs: typing.Unpack[SomeTypedDict]` annotation


nukemiko:
[PEP-692](https://peps.python.org/pep-0692) introduced this machineary for more precise `**kwargs` typing. It seems like cyclopts is not support this.

Consider the following script:

```python
# -*- encoding: utf-8 -*-
from pathlib import Path

from cyclopts import App
from typing import TypedDict
from typing import Unpack


class KwArgs(TypedDict, total=False):
    outputFile: Path
    verbose: bool


app = App()


@app.default
def run(*inputsFiles: Path, **kwargs: Unpack[KwArgs]) -> None:
    print(inputsFiles)
    print(kwargs)


if __name__ == '__main__':
    app()
```

Several outputs are demonstrated that kwargs parsing is totally broken:

```sh-session
$ > python test.py --help
Usage: test.py [OPTIONS] [ARGS...]

╭─ Commands ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --help (-h)  Display this message and exit.                                                                                                                  │
│ --version    Display application version.                                                                                                                    │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Arguments ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ INPUTSFILES                                                                                                                                                  │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Parameters ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --[KEYWORD]                                                                                                                                                  │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
$ > python --verbose
╭─ Error ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Parameter --verbose requires an argument.                                                                                                                    │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
$ > python --verbose 1
╭─ Error ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Invalid value for --verbose: Cannot instantiate typing.Unpack                                                                                                │
[...]

PR #819, opened 2026-05-23
PEP-692 Support (kwargs `Unpack`)

BrianPugh:
Addresses #817 . @nukemiko can you give this a try?


nukemiko:
It seems to have fixed this, thanks! The only regret is I don't have much time to run a fully test.

Issue #820, opened 2026-05-24
Some tests for fish in `test_behavior` fail


PerchunPak:
Looks like CI skips these tests, because it does not install `zsh` and `fish`

```py
=========================== short test summary info ============================
FAILED tests/completion/test_behavior.py::test_behavior[fish-literal-positional] - AssertionError: [literal-positional] expected completions {'foo', 'baz', 'bar'} missing from shell output. partial='poslit command ' got=[]
FAILED tests/completion/test_behavior.py::test_behavior[fish-multi-positional-second] - AssertionError: [multi-positional-second] expected completions {'cat', 'dog'} missing from shell output. partial='multipos command-multi red ' got=[]
FAILED tests/completion/test_behavior.py::test_behavior[fish-equals-form-option-value] - AssertionError: [equals-form-option-value] expected completions {'prod'} missing from shell output. partial='basic deploy --env=p' got=['--env=prod']
FAILED tests/completion/test_behavior.py::test_behavior[fish-multi-positional-third] - AssertionError: [multi-positional-third] expected completions {'small', 'large'} missing from shell output. partial='multipos3 command-multi3 red cat ' got=[]
```

```py
============================= test session starts ==============================
platform linux -- Python 3.13.13, pytest-9.0.3, pluggy-1.6.0
rootdir: /build/source
configfile: pyproject.toml
plugins: cov-nixpkgs-stub-1.0.0, mock-3.15.1, syrupy-5.1.0
collected 2398 items / 2394 deselected / 4 selected                            

tests/completion/test_behavior.py FFFF                                   [100%]

=================================== FAILURES ===================================
____________________ test_behavior[fish-literal-positional] ____________________

scenario_id = 'literal-positional'
app = App(name=('poslit',), alias=(), synonym=(), version_flags=('--version',), help_flags=('--help', '-h'), group=(), validator=[], sort_key=<UNSET>)
prog_name = 'poslit', partial = 'poslit command '
contains = {'bar', 'baz', 'foo'}, excludes = set()
shell_tester_factory = <function fish_tester.<locals>._make_tester at 0x7ffff3df8f40>

    @pytest.mark.parametrize(
        ("scenario_id", "app", "prog_name", "partial", "contains", "excludes"),
        SCENARIOS,
        ids=[s[0] for s in SCENARIOS],
    )
    def test_behavior(
        scenario_id,
        app,
        prog_name,
        partial,
        contains,
        excludes,
        shell_tester_factory,
    ):
[...]

PR #823, opened 2026-05-24
Fix fish completion for positional args and run fish tests in ci

BrianPugh:
addresses #820. @PerchunPak can you give this a try? Fish definitely doesn't get as much attention as bash/zsh. Thanks!


PerchunPak:
Fixes those tests for me, thanks for the quick response!

PR #829, opened 2026-06-02
Harden PyPI publish workflow

shaanmajid:
## Summary

Stacked on top of #828.

- Split the release workflow into separate `build` and `publish` jobs
  - Keeps package building separate from the job that can mint a PyPI trusted publishing token
  - Follows PyPA's recommendation to separate building from publishing
- Scope `id-token: write` to the `publish` job only
  - Ensures OIDC publishing privileges are only available at the point of upload to PyPI
- Pass built distributions through GitHub Actions artifacts
- Disable `setup-uv` caching in the release build path
  - Defense-in-depth measure; avoids building release artifacts from cache state that could be potentially poisoned by other workflow runs

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -7,19 +7,35 @@ on:
       - "v*.*.*"
 
 jobs:
-  publish:
+  build:
     runs-on: ubuntu-latest
     permissions:
-      id-token: write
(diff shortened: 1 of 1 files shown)


BrianPugh:
thanks! we'll tag a release within the next day or two that uses this!

Issue #830, opened 2026-06-08
Automatic short-form aliasing of parameters.


mgielda:
Hi @BrianPugh!

Kudos for continuing to improve Cyclopts. I now asked my team to migrate a mid-sized project from Click to Cyclopts as an experiment. Overall it went pretty smoothly, though one persistent nuisance is that, in order to alias a parameter, you need to do the whole "Annotated" magic.

Example:

```python
    part: Optional[
        Annotated[
            str,
            cyclopts.Parameter(
                alias="-p", help="Specify the FPGA part number"
            ),
        ]
    ] = None,
```

Now I can see that for help, Cyclopts seems to support docstrings (great!). But there is no other way to submit the "-p" alias. I was thinking if solving that problem wouldn't be super awesome as it would let us avoid the whole Annotated nesting and the cyclopts.Parameter etc etc.

Three ways I see to solve this:
- allow to mark a command as automatically generating one letter aliases for all parameters
- add an "alias" decorator like in Feud: https://feud.readthedocs.io/en/latest/sections/decorators/alias.html#feud.decorators.alias
- allow someone to provide one letter abbreviations in the docstrings parameter names somehow, like:

```python
@app.command
def deploy(
    env: Literal["dev", "staging", "prod"],
    replicas: int | Literal["default", "performance"] = "default",
):
    """Deploy code to an environment.

    Parameters
    ----------
    env (e) 
        Environment to deploy to.
    replicas (r)
        Number of workers to spin up.
    """
    if replicas == "default":
        replicas = 10
    elif replicas == "performance":
        replicas = 20

    print(f"Deploying to {env} with {replicas} replicas.")
```

I think having help is important, and the only way to **not** have Annotated **and** have help is to have docstrings. So some sort of docstrings based solution would work best for our use case, I think. But then again, having the "mark command as autoaliasing parameters" would probably be OK too! (or some other solution).

The downside of an explicit "alias" decorator would be that, including docstrings, you would end up listing parameters three times, ouch.

What do you think?

mgielda:
Also, given guidance as to how you'd prefer to approach it, I'd be happy to have my team contribute the implementation.

BrianPugh:
Hey @mgielda!

I agree that the `Annotated` syntax is clunky. In an ideal world, I wish we could do something (but python doesn't allow it) like:

```python
part: Parameter[str, alias="-p"],
```

---

> add an "alias" decorator like in Feud:

[...]

PR #831, opened 2026-06-08
New Attribute: `Parameter.short_alias` to automatically generate short aliases for parameters.

mgielda:
An implementation proposal for https://github.com/BrianPugh/cyclopts/issues/830 - not reviewed yet, but just wanted to elicit comments.

This lets you specify auto_alias=True at either App or Command level - Command overrides App. The auto alias is the first letter of the argument name. If more than 1 arg share the same name, the first one should take precedence. Specific aliases take precedence.

From @BrianPugh 's requests:

* [x] Make linter / static tests pass
* [ ] Align on name
* [ ] Extend Parameter.name and Parameter.alias to allow for a callable that derives names
* [ ] Update docs

### DIFF
--- a/cyclopts/argument/_collection.py
+++ b/cyclopts/argument/_collection.py
@@ -375,9 +375,7 @@ class ArgumentCollection(list[Argument]):
         # Phase 1 here reserves explicit shorts and records eligibility; the actual short is
         # generated in phase 2 (see ``_from_callable``) once every explicit short is known,
         # so an earlier parameter's auto short can't shadow a later parameter's explicit one.
-        if pending_short_aliases is not None and reserve_short_alias(
-            argument, immediate_parameter, used_short_aliases, top_level=not keys
-        ):
+        if pending_short_aliases is not None and reserve_short_alias(argument, immediate_parameter, used_short_aliases):
             pending_short_aliases.append((argument, field_info))
 
         if positional_index is not None:
--- a/cyclopts/argument/_collection.py
+++ b/cyclopts/argument/_collection.py
@@ -480,12 +480,12 @@ class ArgumentCollection(list[Argument]):
                 }
             else:
                 subkey_docstring_lookup = None
-            auto_alias = field_info.kind is field_info.KEYWORD_ONLY or (
+            short_alias = field_info.kind is field_info.KEYWORD_ONLY or (
                 field_info.kind is field_info.POSITIONAL_OR_KEYWORD and not has_star and not field_info.required
             )
             field_parameters: list[Parameter | None] = [
                 Parameter(help=field_info.help) if field_info.help else docstring_lookup.get((field_info.name,)),
-                None if auto_alias else Parameter(auto_alias=False),
+                None if short_alias else Parameter(short_alias=False),
             ]
             iparam_argument_collection = cls._from_type(
                 field_info,
--- a/cyclopts/argument/_collection.py
+++ b/cyclopts/argument/_collection.py
@@ -14,7 +14,6 @@ from cyclopts.exceptions import (
     UnknownOptionError,
 )
 from cyclopts.field_info import (
-    FieldInfo,
     signature_parameters,
 )
 from cyclopts.group import Group
@@ -244,7 +243,7 @@ class ArgumentCollection(list[Argument]):
         docstring_lookup: dict[tuple[str, ...], Parameter] | None = None,
         positional_index: int | None = None,
         used_short_aliases: set[str] | None = None,
-        pending_short_aliases: list[tuple["Argument", FieldInfo]] | None = None,
+        pending_short_aliases: list["Argument"] | None = None,
         _resolve_groups: bool = True,
     ):
         from cyclopts.parameter import get_parameters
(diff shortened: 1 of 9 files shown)


mgielda:
Pushed some updates - now two options starting with the same name get lowercase and uppercase. So only when you hit 3 params starting with the same letter, the third one gets ignored.

BrianPugh:
Thanks for opening this! At a high level, I think the following changes should be made:

1. We need the linter/static-typechecks to pass.
2. I'm not a huge fan of the name `auto_alias`, but I can't think of much better. Maybe `short_alias: bool`? It needs to hit the right case of ambiguous and precision so that we can possibly extend the feature in the future. For example, what if we wanted it to be a Callable so that the user could supply their own name->flag logic?
3. Expanding on (2), maybe we could actually just extend `Parameter.name` and `Parameter.alias` to allow for a callable that derives names. This is kind of inline with other Cyclopts functionality.
4. We need to update the docs. Namely the "API" page and probably a subsection in [Parameter Naming](https://cyclopts.readthedocs.io/en/stable/parameters.html#naming).

So some of these TODOs are pretty open ended, so we should definitely reach a consensus first.

mgielda:
1. Of course!
2. short_alias is fine; I do believe "auto alias" would cover what you mean too, even if the logic is custom it would still be "automatic". But then again I will not die on the naming hill at all :)
3. Not sure yet how that would be done since I don't know Cyclopts internals well enough but I can try to rework the code based on this suggestion.
4. Of course!

So I think we are mostly aligned? Just ACK what name you want to go for and the rest is implementation, which we can handle.

mgielda:
Let's see if my latest commit makes tests green. Will work on the other stuff in parallel.

mgielda:
Attempted 3 in the latest commit - let me know if that is what you had in mind @BrianPugh ?

BrianPugh:
So sorry, i'm just having a realization, could we mostly already achieve this with the current [Parameter.name_transform](https://cyclopts.readthedocs.io/en/stable/api.html#cyclopts.Parameter.name_transform) machinery? We would need to extend it to allow for the return of `str | Iterable[str]`.

We could then do something like:

```python
from cyclopts import default_name_transform, Parameter

def my_custom_name_transform(s: str) -> list[str]:
    original = default_name_transform(s)
    short = your_shorten_function(original)
    return [original, short]

app = App(default_parameter=Parameter(name_transform=my_custom_name_transform))
```

This is getting a bit long-winded/complicated, but it would allow for the greatest flexibility/customization without really enlarging the API surface.

BrianPugh:
ugh, but that could lead to name collisions (which your current implementation doesn't). I need to think about this a little more 🙈

BrianPugh:
@mgielda i've been playing with the implementation of a `Parameter.name_transform` variant, but the primary issue i'm running into there is with dataclass-like objects.

Consider the following:
```python
@dataclass
class User:
    name: str
    color: str


@app.command
 def foo(*, user: User, upload: bool = False):
    ...
```

Under current rules, this naturally generates:
* `--user.name`
*  `--user.color`
* `--upload`

If we were to auto-add short flags, it would also add `-n, -c, -u`. This feels wrong. `--user.n` and `--user.c` also feels wrong. What should our behavior be in these cases?

mgielda:
Maybe `--u.name` `--u.color` and `-U`? Frankly dunno, can you even do short aliases on dataclasses normally (with the alias annotation)? If yes, and if it results in something like `--u.name` then I would say we don't change the logic from if you did it manually. If you can't, then maybe dataclasses just don't autogenerate aliases at all?

mgielda:
@BrianPugh let me know your thoughts on this, and I will have my team take over and implement what is needed.

BrianPugh:
so the more i fiddle with the `name_transform` alternative, the more problems i run into. So perhaps this PR is indeed closer to what we want. No action required on your side yet! i'll just play around with this more. Sorry for the long delay!

BrianPugh:
if you wanna check out the `auto-alias` branch in this repo, i'm refining this PR (and will push the changes to this PR if you agree with it). I'm still in the middle of reviewing it, but basically:

1. Added a lil more testing and docs.
2. **changed it from `auto_alias` to `short_alias`.**
3. Did some forward-searching on already used/declared short flags.
4. Get this working on `Parameter(name="*")` dataclasses.

Posting this now as i'll only have a few hours now-and-then this weekend to keep chipping away at it.

mgielda:
Hi Brian, sure! Just push your commits on top of mine so that it's easy to see what changed, and we will be happy to try out, provide feedback if everything works for us.

And of course thank you! Appreciate your help with this. My goal is to make Cyclopts used across the company, and an easy way to get sane defaults without having to type/think too much I think goes a long way!

BrianPugh:
@mgielda checkout the `auto-alias` branch and merge it into your branch and let me know what you think (i don't have write access to your branch).

The last main change was to be consistent across the codebase, this is only available via `Parameter`, so the global way of defining this is now:

```python
app = App(default_parameter=Parameter(short_alias=True))
```

diogo-rossi:
Hello everyone!

I've made a very small (and limited) package (actually, a single module) named [`clig`](https://clig.readthedocs.io/en/latest/)
that implements a simple rule to generate short alias, which can be used here as an idea.

It can be checked [here](https://clig.readthedocs.io/en/latest/notebooks/advancedfeatures.html#using-make-shorts):

> The strategy to create “short flags” uses the following order:
> - First letter of the argument name (the simplest case)
> - First letter of the argument name UPPERCASED (when there are two names starting with same letter)
> - First letter of the each argument name part if it has TWO_PARTS (separated by underscores)
>
> If the rules above produce ambiguous flags, the strategy starts searching again using first and second lethers. If ambiguity is found again, shearch for first, second and third lethers, and so on.

Thank you all for this great  job!

rw1nkler:
@BrianPugh all the changes work on our end. Thank you!
 I pushed your additions to this PR branch

BrianPugh:
thanks guys! i think i'll probably merge this later today if there are no objections.

PR #833, opened 2026-06-09
Allow `Parameter.show_default` to be a string to display.

### LINKED ISSUE #832: Allow overriding default value in `--help`
I would like to override what is shown in `--help` as the default. In my use case, I define a few `Path` arguments and each of them is resolved later, so I cannot know their value during `--help` invocation. I'd like to just set it to `"automatic"` in the help message

In typer there is a `show_default` argument that accepts a string
https://typer.tiangolo.com/tutorial/options/help/#custom-default-string

This is kind of a duplicate of #458 and #749, but in my use case, I don't use a mutable variable as a default

Example:

```
╭─ Parameters ───────────────────────────────────────────────────────────────────────────╮
│ --input-file -i   The input file with information about entries. [default: automatic]  │
│ --output-file -o  The output file with information about entries. [default: automatic] │
╰────────────────────────────────────────────────────────────────────────────────────────╯
```

BrianPugh:
Resolves #832. @PerchunPak can you give this a try? thanks!

### DIFF
--- a/cyclopts/argument/_argument.py
+++ b/cyclopts/argument/_argument.py
@@ -316,12 +316,14 @@ class Argument:
         return any(dict in (arg, get_origin(arg)) for arg in args)
 
     @property
-    def show_default(self) -> bool | Callable[[Any], str]:
+    def show_default(self) -> bool | str | Callable[[Any], str]:
         """Show the default value on the help page."""
         if self.required:
             return False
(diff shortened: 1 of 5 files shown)


PerchunPak (approved):
Works perfectly, thanks for the fast reply!

Issue #834, opened 2026-06-11
Native slice parsing


pmav99:
Hi Brian! 

I wanted to ask whether native slice support would be in scope for Cyclopts.

My main use case is scientific CLIs that often need to subset dimensions in NetCDF/Zarr-style archives, e.g. time, level etc

So the expected CLI would be like:

```
--time 0:3
--time :10
--time 0:100:5
--time=-10:
```

Cyclopts could validate the syntax directly and return a real slice object. Empty fields map to None.There are probably several stuff that would need to be thought of before making this more concrete but I wanted to float the idea first

BrianPugh:
Added in v4.18.0!

PR #835, opened 2026-06-11
Add slice support

BrianPugh:
Addresses #834. @pmav99 can you give this a try? Thanks!

### DIFF
--- a/cyclopts/_convert.py
+++ b/cyclopts/_convert.py
@@ -176,6 +176,27 @@ def _timedelta(s: str) -> timedelta:
     return timedelta(seconds=seconds)
 
 
+def _slice(s: str) -> slice:
+    """Parse a numpy-style slice string.
+
+    Examples: ``0:3``, ``:10``, ``0:100:5``, ``-10:``. Empty fields map to :obj:`None`.
+
+    Returns
+    -------
+    slice
+    """
+    parts = s.split(":")
+    if 2 <= len(parts) <= 3:
+        try:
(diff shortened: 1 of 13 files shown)


pmav99:
Wow! I did test it and that's absolutely great!!! 

A couple of minor nitpicks. 

1. I am not sure if "numpy-style syntax" is really appropriate for putting in the docs. `3:10:2` is "slice notation". At least to me "numpy style slicing" would point to multi-dimensional slicing as in: 
```python
x[0:100:5, :, -1]
```

2. I am not sure if this should be a hard error, but you are allowed to select an empty slice, e.g. `3:1:1`. Depending on the context this may or may not be desirable. Sometimes you may want to be strict, some other times you may desire the flexibility. Not sure if this can be handled by adding an argument to `Parameter` somehow: `Annotated[slice, Parameter(disallow_empty=True)]`. That's probably too much API though. I would probably leave it as it-is.

BrianPugh:
1. Agreed, for some reason in my mind, numpy-slicing was kind of a superset, but clearly that doesn't make sense as the created object naturally has to be a `slice` 🙈 . I'll update the docstrings.
2. I could add a class to `validators` with an api like:
   ```python
    Annotated[slice, Parameter(validator=validators.Slice(allow_empty=False))]
   ```
   That would then be consistent with other opt-in cyclopts validators and should be pretty easy. I'll follow up with that later today.
   Finally, I could then add a cyclopts convenience type `cyclopts.types.NonEmptySlice` that aliases the above Annotation.

Issue #839, opened 2026-06-19
"Argument middleware"


mo-adamjacobs:
Hi, I'm really enjoying using this library!

I wondered if there's any possibility of something like a command validator that would also let you modify the arguments in the process? Basically a parameter converter, but for all arguments of the command.

The use case I have in mind is, I want my app not to error on missing arguments, but to give the user the opportunity to enter them after hitting enter (like the GitHub CLI for example). An ideal set up to do that would be to allow dropping in a hook that:
- Receives the existing arguments (parsed into the relevant types)
- Receives specs for the remaining arguments to be filled in
- Can access possible completions for the remaining arguments
- Can output modified arguments for the underlying command

Then within the hook I can use `prompt_toolkit` or similar to create a nice interface for filling in the remaining data. This isn't really an ask for Cyclopts to build in that UI (how it's implemented depends on taste) so much as just create space for it via this hook.

Maybe this is something that exists already? Or not quite but there's enough underlying machinery for me to hack it in without changing the source?

Thanks for reading :)

BrianPugh:
Hey @mo-adamjacobs !

This can be accomplished through [the `App.config` system](https://cyclopts.readthedocs.io/en/stable/config_file.html). Here's a minimal example using my preferred prompting library, [questionary](https://github.com/tmbo/questionary) (it uses `prompt_toolkit` under the hood):

```python
import questionary

from cyclopts import App, Token
from cyclopts.argument import ArgumentCollection


def prompt_for_missing(app: App, commands: tuple[str, ...], arguments: ArgumentCollection):
    """Interactively fill in any required argument the user didn't supply."""
    for argument in arguments.filter_by(has_tokens=False):
        if not argument.required:
            continue
        answer = questionary.text(f"{argument.name} ({argument.hint}):").ask()
        if answer is None:  # user hit Ctrl-C
            raise SystemExit(1)
        argument.tokens.append(Token(keyword=argument.name, value=answer, source="interactive"))


app = App(config=prompt_for_missing)


@app.command
def deploy(env: str, replicas: int = 1):
    print(f"Deploying to {env!r} with {replicas} replica(s)")


app()
```

mo-adamjacobs:
Thanks! This works really well and I didn't know about `questionary`. From playing around it seems like this will be good in simple situations, but probably wants custom logic in the command implementation a lot of the time anyway, so a universal solution might not be as easy as I thought.

[...]

PR #840, opened 2026-06-20
Add `ArgumentCollection.filter_by(missing=...)` for required/conditionally-required fields.

BrianPugh:
Simplifies #839 demo to:

```python
import questionary
from datetime import datetime
from cyclopts import App, Parameter, Token
from cyclopts.argument import ArgumentCollection

def prompt_for_missing(app: App, commands: tuple[str, ...], arguments: ArgumentCollection):
    while missing := arguments.filter_by(missing=True):  # re-evaluates after each answer
        argument = missing[0]
        answer = questionary.text(f"{argument.name} ({argument.hint}):").ask()
        if answer is None:  # user hit Ctrl-C
            raise SystemExit(1)
        argument.tokens.append(Token(keyword=argument.name, value=answer, source="interactive"))

@Parameter(name="*")
@dataclass
class TimeRange:
    start: datetime
    end: datetime

app = App(config=prompt_for_missing)

@app.command
def cli(data_source: str, /, *, time_range: TimeRange | None = None):
    print(f"data_source={data_source!r} time_range={time_range!r}")

app()
```

### DIFF
--- a/cyclopts/argument/_argument.py
+++ b/cyclopts/argument/_argument.py
@@ -1297,8 +1297,20 @@ class Argument:
         fully-omitted composite) the checker returns every required child. For multi-branch
         ``Union`` composites, requiredness is branch-aware (see
         :meth:`_active_branch_required_keys`).
+
+        Special case: a *required* multi-branch ``Union`` with no supplied fields hasn't
+        committed to a branch, so :meth:`_active_branch_required_keys` owes nothing — which
+        would leave an interactive prompt loop blind to a value it must collect. Default to
+        the first branch's required fields so the loop converges on (and instantiates) it,
+        matching the declaration-order preference of :meth:`_resolve_union_member`. This
+        read-only query path only; conversion still errors on the composite itself.
         """
         supplied_keys = {child.keys[-1] for child in self.children if child.has_tokens}
+        if self._union_branches and self.required and not supplied_keys:
+            first_branch_required = {k for k, v in self._union_branches[0][1].items() if v.required}
+            return [
+                child for child in self.children if not child.has_tokens and child.keys[-1] in first_branch_required
+            ]
         active_required = self._active_branch_required_keys(supplied_keys)
         if active_required is not None:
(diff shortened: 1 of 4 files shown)


BrianPugh:
@mo-adamjacobs please give this a try!

mo-adamjacobs:
This works nicely for the original example, but I think `Union`s break it, e.g:

```
@dataclass
class TimeRange:
    start: datetime
    end: datetime


@dataclass
class Live:
    live: Annotated[bool, Parameter(negative=())]

@app.command
def cli(data_source: str, /, *, time_period: TimeRange | Live): ...
```

`cli source --time-period.live` -> `Command "cli" parameter --time-period.start requires an argument.`

I'm not sure I could actually find a combination of arguments that gets through here, as supplying them all gives `ValueError: Required field "time_period->origin" is not accessible by Cyclopts; possibly due to conflicting POSITIONAL/KEYWORD requirements.`

I put some thoughts about `Union`s in general in #839 as it gets a bit complicated.

I think this feature seems useful regardless though, and I think could be fixed if a `Union` branch wasn't considered to have missing arguments unless at least one argument on the branch is present? I'm not too sure exactly how this is working yet but it seems to come down to not knowing which child is on which branch. But given how `cyclopts` evaluates `Union`s I guess it is conscious of different branches _somewhere_.

BrianPugh:
@mo-adamjacobs please give it a try now! I added Union support. If the union itself is required, the `filter_by(missing=True)` will use the missing arguments from the first type in the union.

mo-adamjacobs:
Nice, I tested with both `TimeRange | Live` and `Live | TimeRange` and both ways it does what's expected when `--start` is supplied (asks for `--end`). So I think this `missing` filter makes sense. The interaction strategy of "fill in the first missing argument" will currently cause later `Union` branches to be overlooked, but I think I'll need to workshop a full solution to that, and this way works as long as the first one is conceptually "default" which will be a lot of cases.

LGTM, thanks for making the PR!

BrianPugh:
thanks for testing! i think i'll merge this in later today!

PR #843, opened 2026-06-26
Fix the leading-hyphen regression introduced by #639.

### LINKED ISSUE #842: Bug report: `allow_leading_hyphen=True` is broken by #639
Hi, thank you for making and continuously improving cyclopts 👋 ❤️ 

When upgrading cyclopts to the latest version, I noticed a regression when using `allow_leading_hyphen=True` and I bisected it to #639.

Here is a minimal example:

```python
from typing import Annotated
import cyclopts

app = App(help_flags=[], version_flags=[])

@app.default
def k(*args: Annotated[str, Parameter(allow_leading_hyphen=True)]):
    # passthrough wrapper
    print(f"received: {list(args)}")

app(["-c=blue", "build", "thing"])
```

Before #639 (<4.1.0), the output is:

```
received: ['-c=blue', 'build', 'thing']
```

After #639 (>=4.1.0), the output is:

```
received: ['-c', '-=', '-b', '-l', '-u', '-e', 'build', 'thing']
```

`allow_leading_hyphen=True` should take precedence I think? Thanks in advance for taking a look at it!

yilei:
Fixes #842

### DIFF
--- a/cyclopts/bind.py
+++ b/cyclopts/bind.py
@@ -168,6 +168,7 @@ def _parse_kw_and_flags(
                 # Once we hit an option that takes a value, the rest is the value
                 chars = cli_option.lstrip("-")
                 position = 0
+                unmatched_flags: list[str] = []
 
                 while position < len(chars):
                     char = chars[position]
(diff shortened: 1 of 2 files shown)


BrianPugh:
Thank you for opening this! I should be able to get to this on Sunday.

PR #849, opened 2026-07-06
exit 2 on parse errors, `Parameter.negative_alias`, eager `result_action` validation, migration guide

BrianPugh:
## Summary

Final polish on top of `v5-develop` before tagging: two breaking-change cleanups that are cheapest to ship inside a major release (parse-error exit codes, `cyclopts tree -d`), one small feature that closes a `negative=` footgun (`Parameter.negative_alias`), eager validation of `result_action` names, and a "Migrating to v5" docs page covering every intentional breaking change accumulated on the v5 branch.

## Parse errors exit with code 2

CLI usage errors (unknown option, missing argument, coercion failure, ...) now `sys.exit(2)`, matching argparse and Click. Previously they exited `1` — indistinguishable from a command that ran and reported failure, since the default `result_action` maps a `False` return to exit code `1`. Application-level exit codes and `130` for `KeyboardInterrupt` are unchanged.

```console
$ myapp --bogus; echo $?
# before: 1
# after:  2
```

## `--version` respects the `--` delimiter

Version interception now uses the same delimiter-aware flag lookup as help (`_get_help_flag_index` generalized to `_get_flag_index`): a `--version` token after the end-of-options delimiter is positional data, not a version request.

```console
$ myapp -- --version
# before: printed the app version (command never ran)
# after:  the command runs; "--version" is bound positionally
```

## `Parameter.negative_alias`

*Additional* name(s) for a flag's negative form, appended to the generated `--no-*`/`--empty-*` (or `negative`-supplied) names. The `negative`/`negative_alias` pair mirrors `name`/`alias`: `negative` **replaces**, `negative_alias` **appends**. This closes the footgun where `negative="-d"` silently drops the generated `--no-*` long form, and avoids hardcoding the long negative just to add a short one. Primary use-case: a short flag that disables a default-on boolean.

```python
def main(*, description: Annotated[bool, Parameter(negative_alias="-d")] = True): ...
# -d               -> False
# --no-description -> False  (still generated)
# --description    -> True
```

## `cyclopts tree`: `-d` now *disables* descriptions (breaking)

`-d` was a positive alias for `--description` — a no-op, since descriptions default to on. It is now a negative alias (equivalent to `--no-description`), as originally intended, using `negative_alias` above.

## `result_action` names are validated eagerly

An invalid action name now raises a descriptive `ValueError` (listing the valid actions) at `App` construction, attribute assignment, or invocation-time override — **before** any command executes. Previously an invalid name was silently accepted and only raised a bare, message-less `ValueError` at result-handling time, after the command had already run its side effects.

```python
App(result_action="bogus")          # ValueError, raised immediately
app(tokens, result_action="bogus")  # ValueError, command never runs
```

Also fixes the `__call__`/`run_async` parameter docstrings, which claimed the inherited default is `"print_non_int_return_int_as_exit_code"`; the actual fallback (and the documented behavior everywhere else) is `"print_non_int_sys_exit"`.

## "Migrating to v5" docs page

New `docs/source/migration/v5.rst`, linked from the docs index, with before/after examples for each intentional breaking change on the v5 branch: fallthrough child-wins scoping, forwarded `*tokens` preserving the `--` delimiter, greedy `allow_leading_hyphen` `*args` subcommands claiming all post-command tokens, user parameters shadowing `--help`/`--version`, parse-error exit code 2, delimiter-aware `--version`, `cyclopts tree -d`, and the `result_action` validation + default exit-code contract.

### DIFF
--- a/cyclopts/_result_action.py
+++ b/cyclopts/_result_action.py
@@ -1,6 +1,6 @@
 import sys
 from collections.abc import Callable, Iterable
-from typing import Any, Literal, cast
+from typing import Any, Literal, cast, get_args, get_origin
 
 from cyclopts.utils import is_iterable
 
@@ -26,6 +26,25 @@ ResultActionSingle = (
 )
 
 ResultAction = ResultActionSingle | Iterable[ResultActionSingle]
+
+_RESULT_ACTION_NAMES: frozenset[str] = frozenset(
+    get_args(next(arg for arg in get_args(ResultActionSingle) if get_origin(arg) is Literal))
+)
+
+
+def validate_result_action(action: Any) -> None:
+    """Validate a single result action, raising a descriptive :class:`ValueError`.
+
+    Called eagerly (at :class:`App` construction and at invocation-time
+    overrides) so that a typo'd action name fails before any command executes.
+    """
+    if callable(action) or (isinstance(action, str) and action in _RESULT_ACTION_NAMES):
+        return
+    raise ValueError(
+        f"Invalid result_action {action!r}; must be a callable or one of: "
+        + ", ".join(map(repr, sorted(_RESULT_ACTION_NAMES)))
+        + "."
+    )
 
 
 def resolve_returncode(result: Any, default: int = 0) -> int:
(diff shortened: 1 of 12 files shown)


BrianPugh:
Split into independent PRs for cleaner release notes.

PR #858, opened 2026-07-07
Fix Issue 857: pydantic dataclass default factory

### LINKED ISSUE #857: Importing pydantic breaks Dataclass Commands
This works if I invoke the command `test`:

```python
from dataclasses import dataclass, field
#import pydantic

from cyclopts import App

app = App(result_action=["call_if_callable", "print_non_int_sys_exit"])

@app.command
@dataclass
class Test:
    numbers: list[int] = field(default_factory=lambda: [1,2,3])

    def __call__(self) -> str:
        return str(self.numbers)

if __name__ == "__main__":
    app()
```

```shell
python -m test test
[1, 2, 3]
```

If, however, I uncomment the line `import pydantic`, I get:
```shell
python -m test test
╭─ Error ────────────────────────────────────────────────────────────╮
│ 1 validation error for list[int]                                   │
│   Input should be a valid list [type=list_type,                    │
│ input_value=<factory>, input_type=_HAS_DEFAULT_FACTORY_CLASS]      │
│     For further information visit                                  │
│ https://errors.pydantic.dev/2.13/v/list_type                       │
╰────────────────────────────────────────────────────────────────────╯
```

BrianPugh:
Fixes #857.

## Root cause

When a class is registered directly as a command, its parameters are derived from `inspect.signature(cls)` — and the raw `__init__` signature uses library-private sentinels as the "default" for factory fields:

| Library | Sentinel in `__init__` signature |
|---|---|
| `dataclasses` | `_HAS_DEFAULT_FACTORY` |
| pydantic `BaseModel` | `_HAS_DEFAULT_FACTORY` (reused by pydantic's generated `__signature__`) |
| attrs | `attrs.NOTHING` |

With pydantic installed, cyclopts validates argument defaults through `pydantic.TypeAdapter.validate_python`, which rejects these sentinels — so merely `import pydantic` broke otherwise-working dataclass commands. The sentinels also leaked into help text as `[default: <factory>]` / `[default: _Nothing.NOTHING]`.

Investigating the issue surfaced the same bug family in sibling paths, all fixed here (each with a regression test written and confirmed failing first):

## Fixes

- **Dataclass as command** (the original report): `default_factory` sentinel validated by pydantic; help showed `<factory>`.
- **attrs class as command**: same failure via the `attrs.NOTHING` sentinel; help showed `_Nothing.NOTHING`.
- **pydantic `BaseModel` as command**: same failure via pydantic's reuse of the dataclasses sentinel.
- **Nested attrs field with a factory**: `_attrs_field_infos` substituted `default=None` for factory fields, which pydantic then rejected against the real annotation (e.g. `list[int]`).
- **`Parameter(accepts_keys=True)` on a dataclass with a `default_factory` field**: raised a bare `NotImplementedError` — the field was registered twice (resolved default vs. raw sentinel) and the mismatch hit `_update_lookup`'s fallback. The `__init__`-signature fall-through now skips already-registered fields while retaining its real jobs (`**kwargs` capture, unrecognized types).

## Design

`signature_parameters` now resolves class defaults with a single merge of `get_field_infos()` values over the raw signature, so each library's factory policy lives in exactly one place (its `_*_field_infos` helper). That policy is now **consistent across dataclasses, attrs, and pydantic**: invoke the factory during introspection and use the result as the default. Factories that cannot be invoked without an instance — attrs `takes_self=True` and pydantic data-taking factories (`lambda data: ...`, pydantic ≥2.10) — are treated as having no introspectable default (nothing shown in help, nothing validated).

## User-visible behavior changes

Beyond the crash fixes, factory defaults now render in help text where they previously showed nothing (or a sentinel):

```text
╭─ Parameters ───────────────────────────────────────╮
│ NUMBERS --numbers  [default: [1, 2, 3]]            │
╰────────────────────────────────────────────────────╯
```

This applies to nested attrs fields and pydantic `default_factory` fields (dataclasses already behaved this way). Note that factories are now invoked at introspection time for attrs (they already were for dataclasses), so factories with side effects will run during help generation/parsing. Docs snapshots and expected help strings updated accordingly.

## Testing

- 7 new regression tests across `test_pydantic.py`, `test_bind_attrs.py`, and `test_bind_dataclasses.py`, each confirmed failing before its fix.
- Full suite passes (2536 passed); snapshot diffs audited to contain only the new `[default: ...]` annotations.

### DIFF
--- a/cyclopts/argument/_argument.py
+++ b/cyclopts/argument/_argument.py
@@ -275,7 +275,10 @@ class Argument:
                     continue
                 if field_info.kind is field_info.VAR_KEYWORD:
                     self._default = field_info.annotation
-                else:
+                elif field_info.name not in self._lookup:
+                    # Fields already registered via ``get_field_infos`` have richer metadata
+                    # (e.g. resolved ``default_factory`` values) than the raw ``__init__``
+                    # signature; don't re-register them.
                     self._update_lookup({field_info.name: field_info})
 
         if self._accepts_keywords and len(hints) > 1:
--- a/cyclopts/field_info.py
+++ b/cyclopts/field_info.py
@@ -363,34 +363,15 @@ def signature_parameters(f: Any) -> dict[str, FieldInfo]:
         annotation = type_hints.get(name, iparam.annotation)
         out[name] = FieldInfo.from_iparam(iparam, annotation=annotation)
 
-    if inspect.isclass(func) and is_dataclass(func):
-        # A dataclass's ``__init__`` signature uses a private sentinel as the default
-        # for ``default_factory`` fields; resolve it to the actual factory value.
-        import dataclasses
-
-        for dataclass_field in dataclasses.fields(func):
-            if dataclass_field.default_factory is not dataclasses.MISSING and dataclass_field.name in out:
-                out[dataclass_field.name] = out[dataclass_field.name].evolve(default=dataclass_field.default_factory())
-    elif inspect.isclass(func) and is_attrs(func):
-        # An attrs class's ``__init__`` signature uses the ``attrs.NOTHING`` sentinel as the
-        # default for factory fields; resolve it to the actual factory value.
-        # ``takes_self`` factories cannot be invoked without an instance; treat those
-        # fields as having no introspectable default.
-        for attribute in func.__attrs_attrs__:
-            if isinstance(attribute.default, attrs.Factory) and attribute.alias in out:  # pyright: ignore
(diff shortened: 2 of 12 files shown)


BrianPugh:
@clo-vis can you test this? Thanks!

clo-vis:
> @clo-vis can you test this? Thanks!

Sorry for the late reply (I was on vacation). Yes, it works perfectly - but you already knew that.

PR #860, opened 2026-07-14
fix: complete zsh subcommands after meta positionals

### LINKED ISSUE #759: Meta app arguments affect shell completion
Hello, first of all thank you for such a fantastic library.

I've written a CLI using meta apps that injects positional arguments between commands, in other words I have subcommands that branch off positional arguments. For example:

*vban-cli strip <strip_index> eq cell <cell_index> on true*

```console
vban-cli strip 0 eq cell 3 on true
```

Unfortunately, this affects tab completion once I enter the index values.

Let's say I enter `vban-cli strip` and press TAB, it shows me all the subcommands, however, if I enter `vban-cli strip 0` and press TAB, I get no output, even though the strip_index is required by the CLI and will fail if not entered.

---

I suspect the solution requires reworking the completion logic to analyse arguments in meta apps. The completion script would need to know that a positional argument exists in place of a subcommand.

Sanjays2402:
Fixes #759.

Zsh always placed a nested command in positional slot 1. A required positional on a meta app occupied that slot, so `vban strip 0 <TAB>` could no longer enter the subcommand state.

Generate only the positionals introduced at the current path and shift command completion/dispatch after fixed positional-only arguments. The regression test drives a real zsh session and verifies that `mute` appears after the required index.

Tests: `pytest -q tests/completion` (198 passed, 49 skipped)

### DIFF
--- a/cyclopts/completion/_base.py
+++ b/cyclopts/completion/_base.py
@@ -10,6 +10,8 @@ from enum import Enum
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, get_args, get_origin
 
+from attrs import field
+
 from cyclopts.annotations import ITERABLE_TYPES, is_annotated, is_union
 from cyclopts.argument import ArgumentCollection
 from cyclopts.exceptions import CycloptsError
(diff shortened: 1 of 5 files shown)


BrianPugh:
Thanks for the PR! i'm gonna run copilot on this just for my own reference; don't feel obligated to respond/interact with it

BrianPugh:
@Sanjays2402 can you give this a try?

Sanjays2402:
@BrianPugh I tried the current head with real zsh 5.9 on macOS. `vban strip 0 <TAB>` expands to `vban strip 0 mute `, while `vban strip <TAB>` correctly does not offer `mute` before the required index.

With the local temp-path fixture quoted (my SSD test root contains a space), the full completion suite is green: **205 passed, 49 skipped**. The branch itself is unchanged.

BrianPugh:
awesome thanks! i'll merge this in soon!

PR #861, opened 2026-07-17
Fix `--` delimiter leaking into an "Unknown option" error

chuenchen309:
## The bug

When a `POSITIONAL_OR_KEYWORD` parameter is passed as a keyword and a later positional value follows the `--` end-of-options delimiter, cyclopts raises the wrong error — it reports the `--` delimiter itself as an unknown option:

```python
from cyclopts import App
app = App()

@app.default
def main(x: str, rest: list[str] = []): ...

app(["--x", "val", "--", "pos"], exit_on_error=False)
# UnknownOptionError: Unknown option: --. Did you mean --x?
```

Without the delimiter, the same input gives the correct, meaningful diagnosis:

```python
app(["--x", "val", "pos"], exit_on_error=False)
# ArgumentOrderError: Cannot specify token "pos" positionally for parameter rest
#                     due to previously specified keyword --x. ...
```

The `--` delimiter only forces the tokens after it to be positional; it must not change the diagnosis. So `--x val -- pos` should behave exactly like `--x val pos`.

## Cause & fix

`_parse_pos` (`cyclopts/bind.py`) read the raw `tokens[0]` in this branch — which still contains the `--` delimiter — and ran `is_option_like` on it, so `--` was reported as an unknown option. The sibling paths already use the preprocessed `tokens_and_force_positional` list (delimiter stripped, each token carrying a `force_positional` flag): the `ArgumentOrderError` token just below it, and the option-like check in the consume loop. This one branch was the lone spot still reading raw `tokens`.

The fix reads from `tokens_and_force_positional` and honors `force_positional`, so a post-`--` token falls through to `ArgumentOrderError`, while a genuine leading-hyphen option (`--x val -foo`) still raises `UnknownOptionError("-foo")`. Verified `--x val -- pos`, `--x val -- -x a`, and `--x val -foo` all now behave correctly.

## Tests

Added a regression test next to `test_out_of_order_mixed_positional_or_keyword` (TDD, as the contributing guide suggests: it is red before the fix — `UnknownOptionError` — and green after). `tests/test_bind_basic.py` passes (138), and `ruff check`, `ruff format --check`, and `pyright` (strict) are all clean.

---
*Disclosure: prepared with AI assistance (Claude Code, Claude Opus 4.8). I reproduced the bug, wrote the failing regression test first, verified the fix against neighboring input cases and the strict lint/type gates, and reviewed the diff before opening.*

### DIFF
--- a/cyclopts/bind.py
+++ b/cyclopts/bind.py
@@ -448,6 +448,10 @@ def _parse_pos(
                 # Continue in case we hit a VAR_POSITIONAL argument.
                 continue
             if prior_positional_or_keyword_supplied_as_keyword_arguments:
+                if not tokens_and_force_positional:
+                    # ``tokens`` contained only the ``--`` end-of-options delimiter;
+                    # there are no positional tokens to misassign.
+                    break
                 # Use the preprocessed token: ``tokens`` still contains the ``--``
                 # end-of-options delimiter, and ``force_positional`` marks tokens
                 # after it, which must not be treated as options.
--- a/cyclopts/bind.py
+++ b/cyclopts/bind.py
@@ -448,8 +448,11 @@ def _parse_pos(
                 # Continue in case we hit a VAR_POSITIONAL argument.
                 continue
             if prior_positional_or_keyword_supplied_as_keyword_arguments:
-                token = tokens[0]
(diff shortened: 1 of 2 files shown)


BrianPugh:
thank you!

Issue #370, opened 2025-03-18
Config search_parents=True not working when config path provided as relative path


robmcd:
https://github.com/BrianPugh/cyclopts/blob/7dbb85141b085696043d41ad7d6dd0e2d0017dbd/cyclopts/config/_common.py#L179

I wonder if this line should be:

`for parent in self.path.resolve().parents:` (added `.resolve()`)

I have an issue where when I specify my config path as `config.toml` and the file is in a parent directory of the working directory, the config is not found.

During debugging I see:

self.path
Out[1]: WindowsPath('config.toml')

list(self.path.parents)
Out[2]: [WindowsPath('.')]

list(self.path.resolve().parents)
Out[3]: 
[WindowsPath('C:/a/b/c/d/e/f/g/h'),
 WindowsPath('C:/a/b/c/d/e/f/g'),
 WindowsPath('C:/a/b/c/d/e/f'),
 WindowsPath('C:/a/b/c/d/e'),
 WindowsPath('C:/a/b/c/d'),
 WindowsPath('C:/a/b/c'),
 WindowsPath('C:/a/b'),
 WindowsPath('C:/a'),
 WindowsPath('C:/')]

IF I specify the full path to my config it works.

BrianPugh:
Thanks for the report! Fixed in v3.10.1 which should be out momentarily.

PR #864, opened 2026-07-19
Fix search_parents=False still walking parent directories

chuenchen309:
### The bug

With the default `search_parents=False`, a config whose `path` doesn't exist still loads a same-named config file from an **ancestor** directory:

```python
import cyclopts
from cyclopts import App

# /tmp/base/pyproject.toml exists with [tool.myapp.count] character = "PARENT_VALUE"
# /tmp/base/sub/pyproject.toml does NOT exist
app = App(name="myapp", config=cyclopts.config.Toml(
    "/tmp/base/sub/pyproject.toml", root_keys=["tool", "myapp"]))
# search_parents defaults to False, must_exist defaults to False

@app.command
def count(filename: str, *, character: str = "DEFAULT"):
    return character

app(["count", "README.md"], result_action="return_value")
# Actual:  'PARENT_VALUE'  (loaded from the ancestor)
# Expected: 'DEFAULT'      (search_parents=False must not consult ancestors)
```

The documented contract (`docs/source/api.rst`, and the primary example in `docs/source/config_file.rst`) makes parent search opt-in: the attribute defaults to `False`, and every example that wants ancestor lookup sets `search_parents=True` explicitly. So with the default, only the specified `path` should be consulted.

Impact: any tool using the headline `config=cyclopts.config.Toml("pyproject.toml", ...)` pattern (default flags) silently absorbs the repo-root `pyproject.toml` when run from a subdirectory, even though the user never opted into parent search.

### Root cause

`ConfigFromFile.config` (`cyclopts/config/_common.py`) walks `self.path...parents`. When a candidate doesn't exist, only two cases were handled:

```python
elif self.search_parents:
    continue
elif self.must_exist:
    raise FileNotFoundError(...)
```

With both `search_parents=False` and `must_exist=False` (the default), neither branch runs, so the `for` loop simply advances to the next parent — the flag is inert. It only ever had an effect because `must_exist=True` happened to raise early.

### The fix

`break` out of the loop after the specified path unless `search_parents` is set. `must_exist` is still honored by the existing post-loop check, so all four flag combinations behave as documented:

- `search_parents=False` (default) → only the given path; missing → `{}` (or `FileNotFoundError` if `must_exist=True`).
- `search_parents=True` → walk ancestors as before.

### Verification

- The repro above now returns `'DEFAULT'`; `search_parents=True` still returns `'PARENT_VALUE'`; `must_exist=True` on a missing path still raises `FileNotFoundError`.
- Added `test_config_common_search_parents_false_ignores_parent`: it fails on `main` (the ancestor config is loaded) and passes with the fix.
- `tests/config/` → 109 passed. Full suite → 2395 passed, 139 skipped (the only failures are pre-existing `test_docs_snapshots`/`test_sphinx_ext` cases that need `docutils`, which isn't installed here — they fail identically without this change). `ruff check` / `ruff format --check` clean.

This is a behavior change for the default configuration, so it's your call whether it should land as a fix or wait for a major version — happy to adjust the framing. It's distinct from #370 (that was `search_parents=True` failing for *relative* paths, fixed in v3.10.1 by adding `.resolve()`); this is the default `False` not being honored.

---

This PR was authored by an AI coding agent (Claude Code) running on this account: the AI found the bug, ran the repro, wrote the test, and wrote this description. The human account holder reviews every change and is accountable for it. The verification above is real and re-runnable from the diff. If this isn't the kind of contribution you want, say so and I'll close it.

### DIFF
--- a/cyclopts/config/_common.py
+++ b/cyclopts/config/_common.py
@@ -160,11 +160,9 @@ class ConfigFromFile(ConfigBase):
                         msg += exception_msg
                     raise CycloptsError(msg=msg) from e
                 return self._config
-            elif self.search_parents:
-                # Continue iterating over parents.
-                continue
-            elif self.must_exist:
-                raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT), str(self.path))
+            if not self.search_parents:
+                # Only look at the specified path; do not walk parent directories.
+                break
 
         # No matching file was found.
         if self.must_exist:
--- a/tests/config/test_common.py
+++ b/tests/config/test_common.py
@@ -157,8 +157,7 @@ def test_config_common_search_parents_absolute_true_exists(tmp_path, must_exist,
 
 
 def test_config_common_search_parents_false_ignores_parent(tmp_path, config, mocker):
-    """With the default ``search_parents=False``, a missing path must not fall
-    back to a same-named config file in a parent directory."""
+    """A missing path must not fall back to a parent's same-named config when ``search_parents=False``."""
     spy_load_config = mocker.spy(config, "_load_config")
 
     original_path = config.path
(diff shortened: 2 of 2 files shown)


BrianPugh:
thank you!

PR #869, opened 2026-07-22
Fix Literal types containing a None member

BrianPugh:
## Problem

A `Literal` annotation containing a `None` member crashed on valid input: for `Literal[None, "foo"]`, passing `foo` raised an uncaught `TypeError: NoneType takes no arguments`, because the Literal conversion loop tries `convert(type(choice), token)` for each member left-to-right and only `CoercionError` is caught.

`Optional[Literal["foo"]]` already worked correctly (the Union path skips the `None` member), so the equivalent `Literal[None, "foo"]` spelling should behave the same.

## Fix

Treat a `None` Literal member like `Optional`: reachable only via the default value, never via a token. The `None` member leaked in three places, all fixed to match:

- `cyclopts/_convert.py`: skip `None` choices during token conversion (the crash itself).
- `cyclopts/exceptions.py`: omit `None` from "Choose from:" coercion-error messages.
- `cyclopts/argument/utils.py` (`get_choices_from_hint`): omit `None` from help/completion choices, so help renders `[choices: fizz, buzz]` instead of including a spurious `None`.

## Tests

- `test_coercion.py::test_coerce_literal_none_member` — valid tokens convert regardless of `None` member position.
- `test_coercion.py::test_coerce_literal_none_member_invalid_choice` — error message lists only enterable choices.
- `test_help.py::test_help_format_group_parameters_choices_literal_none_member` — help choices omit `None`.

No behavior change for Literals without a `None` member.

### DIFF
--- a/cyclopts/_convert.py
+++ b/cyclopts/_convert.py
@@ -658,6 +658,9 @@ def _convert(
         # Try coercing the token into each allowed Literal value (left-to-right).
         last_coercion_error = None
         for choice in get_args(type_):
+            if choice is None:
+                # Like ``Optional``, a ``None`` member is only reachable via the default, not a token.
+                continue
             try:
                 res = convert(type(choice), token)
             except CoercionError as e:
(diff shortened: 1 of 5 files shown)


BrianPugh:
closing as a pretty niche case that is inherently better handled in the upcoming v5.

PR #888, opened 2026-08-13
Fix inconsistent whitespace after parameter colon in markdown docs (#887)

### LINKED ISSUE #887: Inconsistent whitespace after param colon in generated docs
`generate_docs()` produces varying amounts (0-3) of whitespace after the colon for a parameter, depending on whether the param has help, metadata or is in the last position.

For example:

```python
from cyclopts import App

app = App(name="something")


@app.command
def do_something(
    *,
    has_help: str | None = None,  # ✅ one space
    has_metadata: str,  # ❌ three spaces
    has_no_help: str | None = None,  # ❌ one trailing space
    is_last: str | None = None,  # ✅ no trailing space
):
    """Do something.

    Args:
        has_help: Some help text.
    """


print(app.generate_docs(), end="")
```

->

```markdown
...
**Parameters**:

* `--has-help`: Some help text.
* `--has-metadata`:   **[required]**
* `--has-no-help`: 
* `--is-last`:
```

I would assume what we'd actually want is always one space, unless it's trailing (i.e. there's no further text after the param name).

i.e.

```diff
**Parameters**:
 
 * `--has-help`: Some help text.
-* `--has-metadata`:   **[required]**
-* `--has-no-help`: 
+* `--has-metadata`: **[required]**
+* `--has-no-help`:
 * `--is-last`:
```

For the no-further-text, I guess we'd ideally also want no colon?

The reason I noticed this is that we use a unit test to autogenerate docs and assert they're unchanged, which will always fail if using a pre-commit hook which strips trailing whitespace (i.e.  [`trailing-whitespace`](https://github.com/pre-commit/pre-commit-hooks#trailing-whitespace)), necessitating manual pre-stripping.

BrianPugh:
Fixes #887.

## Problem

`generate_docs()` produced varying amounts of whitespace (0–3 spaces) after a parameter's colon in the **markdown** output, depending on whether the parameter had help text, metadata, or was in the trailing position:

```markdown
* `--has-help`: Some help text.
* `--has-metadata`:   **[required]**
* `--has-no-help`: 
* `--is-last`:
```

The trailing space on parameters with no following text also breaks pre-commit's [`trailing-whitespace`](https://github.com/pre-commit/pre-commit-hooks#trailing-whitespace) hook when docs are autogenerated and checked in.

## Fix

In `cyclopts/help/formatters/markdown.py`, write a bare colon and let each following piece (description, `**[required]**`, and each `*[…]*` metadata item) supply its own single leading space. There is now always exactly one space when text follows the colon, and no trailing whitespace otherwise:

```markdown
* `--has-help`: Some help text.
* `--has-metadata`: **[required]**
* `--has-no-help`:
* `--is-last`:
```

Scope is markdown-only — RST (definition-list layout) and HTML (metadata spans) never exhibited the issue.

## Tests

- Added `test_generate_docs_consistent_whitespace_after_param_colon` reproducing the issue, including a blanket assertion that no rendered line has trailing whitespace.
- Updated existing hardcoded expectations in `test_generate_docs.py` and `test_markdown_formatter.py`, and regenerated the syrupy markdown snapshots. The snapshot diff is whitespace-only (verified: every changed line is identical after collapsing spaces).
- Full suite: 2636 passed, 6 skipped.

### DIFF
--- a/cyclopts/help/formatters/markdown.py
+++ b/cyclopts/help/formatters/markdown.py
@@ -156,8 +156,11 @@ class MarkdownFormatter:
                     else:
                         name_str = ", ".join(long_opts)
 
-                # Start the entry (no type display)
-                self._output.write(f"* `{name_str}`: ")
+                # Start the entry (no type display). No trailing space after the
+                # colon; each following piece (description/metadata) supplies its
+                # own single leading space so there is never trailing whitespace
+                # or multiple spaces after the colon (see #887).
+                self._output.write(f"* `{name_str}`:")
 
                 # Add description with proper indentation for nested content
                 desc = extract_text(entry.description, console, preserve_markup=True)
@@ -166,7 +169,7 @@ class MarkdownFormatter:
 
                     # Split into lines and indent continuation lines to nest under the bullet
                     lines = desc.split("\n")
-                    self._output.write(lines[0])  # First line on same line as bullet
+                    self._output.write(" " + lines[0])  # First line on same line as bullet
 
                     # Track what type of list context we're in for proper nesting
                     in_numbered_list = False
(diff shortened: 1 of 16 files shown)


BrianPugh:
@stainbank can you confirm this fixes it on your end? Thanks!

stainbank:
Looks perfect -- thanks @BrianPugh

PR #892, opened 2026-08-19
Report an un-parsable Union as a clean error instead of a traceback

BrianPugh:
## Summary

`token_count` raises a plain `ValueError` when a Union's members consume differing numbers of tokens. It is called from `Argument.token_count` during token binding, which is outside Cyclopts's error handling, so annotations like `Union[list[int], int]` and `Union[bool, str]` gave the end user a raw Python traceback and silently bypassed `exit_on_error`.

Nothing surfaces at registration or in `--help`, so the app looks healthy right up until someone actually passes the option.

## Fix

`Argument.token_count` now translates that `ValueError` into a `CycloptsError` carrying the offending argument, so it is formatted like any other CLI error and honors `exit_on_error`. `token_count` itself still raises `ValueError` as before, keeping its contract (and existing test) intact; its message now names the types readably rather than printing raw `repr`s.

## Testing

Adds `tests/test_bind_union.py` cases asserting a `CycloptsError` (not a raw traceback) and that `exit_on_error` is honored. The 5 new tests fail on `main` before this change and pass after.

### DIFF
--- a/cyclopts/_convert.py
+++ b/cyclopts/_convert.py
@@ -27,6 +27,7 @@ else:
 from cyclopts.annotations import (
     ITERABLE_TYPES,
     get_annotated_discriminator,
+    get_hint_name,
     is_annotated,
     is_enum_flag,
     is_nonetype,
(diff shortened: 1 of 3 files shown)


BrianPugh:
moot in v5

PR #893, opened 2026-08-19
Raise a clean AmbiguousCommandError instead of a raw ValueError traceback

BrianPugh:
## Summary

Two commands colliding under fuzzy name normalization (e.g. `my-command` and `my_command` with input `mycommand`) crashed `app()` with an unhandled `ValueError` traceback: the error was raised from command resolution inside `app_stack.__enter__`, before `parse_args`' `CycloptsError` handling, so it bypassed error formatting and ignored `exit_on_error`.

## Fix

The ambiguity now raises `AmbiguousCommandError` (a `CycloptsError`; still a `ValueError` subclass for backwards compatibility) through the normal error path: formatted panel, `exit_on_error`/`print_error` honored. The `app_stack` resolution pass uses the new `parse_commands(raise_on_ambiguous=False)` since it only feeds override propagation.

## Testing

Adds `tests/test_fuzzy_command_matching.py` coverage. Verified on `main` that `app(["mycommand"], exit_on_error=False)` currently escapes with a raw `ValueError`; after this change it raises a formatted `AmbiguousCommandError`. `AmbiguousCommandError` is exported and documented in `api.rst`.

### DIFF
--- a/cyclopts/__init__.py
+++ b/cyclopts/__init__.py
@@ -7,6 +7,7 @@ except ImportError:
 
 __all__ = [
     "__version__",
+    "AmbiguousCommandError",
     "App",
     "Argument",
     "ArgumentCollection",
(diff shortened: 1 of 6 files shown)


BrianPugh:
Closing: on v5-develop this no longer crashes (already resolves to a clean UnknownCommandError), so only a low-value semantic rename to AmbiguousCommandError would carry forward. Not worth the effort.

PR #894, opened 2026-08-19
Fix Literal types containing a None member

BrianPugh:
## Summary

`Literal[None, "a", "b"]` fails whenever the `None` choice is reached during conversion. On `main`, passing the **valid** value `b` is rejected with `CoercionError: NoneType takes no arguments`, because conversion tries the `None` member and crashes before reaching the real choice. Invalid input surfaces the same internal `NoneType takes no arguments` text.

## Fix

Skip `None` members during token conversion. Like `Optional`, `None` is only reachable via the default, never via a token. Also omit `None` from "Choose from:" error messages, help-panel choices, and shell-completion choices, matching the equivalent `Optional[Literal[...]]` spelling.

## Testing

Adds coercion and help-panel coverage for `None`-containing `Literal`s. New tests fail on `main` before this change and pass after.

### DIFF
--- a/cyclopts/_convert.py
+++ b/cyclopts/_convert.py
@@ -658,6 +658,9 @@ def _convert(
         # Try coercing the token into each allowed Literal value (left-to-right).
         last_coercion_error = None
         for choice in get_args(type_):
+            if choice is None:
+                # Like ``Optional``, a ``None`` member is only reachable via the default, not a token.
+                continue
             try:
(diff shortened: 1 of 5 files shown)


BrianPugh:
Closing: moot on v5-develop (v5 already accepts valid Literal[None,...] values and gives clean errors); only a cosmetic 'omit None from choices' delta remains. Not worth the effort.

PR #899, opened 2026-08-22
Add type hints to help copy methods

Ichunjo:
Hello.

I realized these helper methods were untyped when trying to customize the `HelpEntry` which forced me to add `# type: ignore[no-untyped-call]` to the `.copy()`.


BrianPugh:
Thank you!

PR #900, opened 2026-08-25
Fix deprecated not being rendered

wallento:
The deprecated directive is not being rendered. The reason is that it is split out by docstring as its own metaclass. Also retrieve this when formatting the doc and add to the components.

Also pull out the formatting, and allow empty deprecation version.

### DIFF
--- a/cyclopts/help/help.py
+++ b/cyclopts/help/help.py
@@ -363,10 +363,18 @@ def format_doc(app: "App", format: str) -> InlineText | SilentRich:
     if parsed.short_description:
         components.append(parsed.short_description + "\n")
 
+    if parsed.deprecation and format in ("restructuredtext", "rst"):
+        from cyclopts.help.rst_preprocessor import format_deprecated_tag
+
+        if components:
+            components.append("\n")
+        components.append(format_deprecated_tag(parsed.deprecation.version, parsed.deprecation.description) + "\n")
+
     if parsed.long_description:
-        if parsed.short_description:
+        if components:
             components.append("\n")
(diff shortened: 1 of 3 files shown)


BrianPugh:
@wallento let me know what you think of these changes.

Extending the conversation from #901, I would be open to both an `App.deprecated` (in your use-case, would be used as `@app.command(deprecated: str =...)`) as well as (for consistent API) `Parameter.deprecated`. Maybe it should be a `str | tuple[str, str]` (for also indicating the version it was deprecated)? We should probably have that be a separate PR for this (and an app/parameter explicit field should override the docstring-inferred `deprecated`).

> hides the entry from the app help?

I don't think we should automatically set that if deprecated, I think it's better for them to additionally explicitly set `App.show=False`.

wallento:
Awesome, thank you. I can implement #901 then, and I agree about the version, thanks.

One thing to consider is another minor addition: If something is deprecated and will be removed, another lifecycle could be when it is scheduled to be removed. But as Sphinx does not have it, I would suggest to not add `.. scheduledremove::` or so.

Oh, right, thanks, I did not see `show` before, that is very helpful.

This PR is good for me then, I will open another one for #901

BrianPugh:
thanks! i'll hold off on releasing so that we can merge in your upcoming PR (and then do a release)

wallento:
~One quick question, maybe out of ignorance. Is `tupel[str,str]` deprecation and removal? Or is it `str` message and `tuple[str,str]` message and version?~ sorry, got it

PR #902, opened 2026-08-25
Split JSON documents on newlines when rendering decode errors

dylanpulver:
`json_decode_error_verbosifier` splits the document with `str.splitlines()` and indexes it with `JSONDecodeError.lineno`. The two disagree on what a line is. `lineno` comes from `doc.count("\n", 0, pos) + 1`, while `splitlines()` also breaks on `\v`, `\f`, `\x1c`-`\x1e`, `\x85`, `\u2028`, `\u2029`, and drops the empty final element.

`\u2028` and `\u2029` are accepted raw inside a JSON string by `json.loads`, so a document containing one yields more entries from `splitlines()` than `lineno` counts. The caret is then drawn under a line other than the failing one, no error raised.

When the failure sits at the end of a document ending in a newline, `lineno - 1 == len(lines)` and the lookup raises `IndexError`. An empty `config.json` reaches it. The `IndexError` escapes from `CoercionError.__rich__` while rich renders the error panel, so a rich-internal traceback replaces the message.

`split("\n")` is how `lineno` is derived, so the index is in range by construction: `len(doc.split("\n")) == doc.count("\n") + 1 >= lineno`. Bounds-checking `splitlines()` would stop the `IndexError` but leave the wrong-line case, so I did not take that route.

Unchanged: `colno` is a character offset, so with tabs or double-width characters the caret is off by the display width.

Five new assertions cover both failure modes, one through an empty config file end to end. Each fails on `main`. Suite goes 2614 to 2619 passed with no other change; `pre-commit run --all-files` is clean.

### DIFF
--- a/cyclopts/utils.py
+++ b/cyclopts/utils.py
@@ -404,7 +404,11 @@ def json_decode_error_verbosifier(decode_error: "JSONDecodeError", context: int 
     context: int
         Number of surrounding-character context
     """
-    lines = decode_error.doc.splitlines()
+    # ``JSONDecodeError.lineno`` is derived from ``doc.count("\n", 0, pos) + 1``, so the document
+    # has to be split on ``"\n"`` to stay in step with it. ``str.splitlines()`` additionally breaks
+    # on ``\v``, ``\f``, ``\u2028`` and friends, which JSON allows raw inside strings, and it drops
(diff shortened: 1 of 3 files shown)


BrianPugh (approved):
thank you! this will be released sometime within the next few days!

PR #903, opened 2026-08-25
Add structured command/parameter deprecation with runtime warning hook

wallento:
Command/Parameter previously only supported deprecation via docstring `.. deprecated::` directives. Adds an explicit `deprecated` field (message or (version, message) tuple) that takes priority over the docstring, plus a `deprecated_handler` callable fired at actual point-of-use (command invocation, parameter explicitly supplied) so callers can route notices through their own logger instead of relying on Python's warnings filter. CommandSpec.deprecated is now wired through to the resolved App.

### DIFF
--- a/cyclopts/_deprecation.py
+++ b/cyclopts/_deprecation.py
@@ -1,0 +1,27 @@
+import warnings
+from collections.abc import Callable
+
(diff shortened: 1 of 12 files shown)


wallento:
Draft still, needs a bit more of review by me. I chose to have a hook for logger callbacks, so that warnings are not automatically logged

PR #905, opened 2026-08-26
Render flattened subapp commands in generated docs

### LINKED ISSUE #904: Commands in flattened subapps don't fully render in generated docs
The cyclopts [subcommand docs](https://cyclopts.readthedocs.io/en/latest/commands.html#flattening-subcommands) suggest using `app.command(sub_app, name="*")` to flatten a subapp into its parent command.

However `generate_docs` does not render the command help completely.

By contrast, using `app.update(sub_app)` (which is undocumented?) seems to achieve the same result, and _does_ render the flattened command properly.

For example:

```python
from cyclopts import App

app = App(name="app")

sub_app_update = App(name="update")
sub_app_star = App(name="star")


@sub_app_update.command()
def update(foo: int) -> None: ...


@sub_app_star.command()
def star(bar: int) -> None: ...


app.command(sub_app_star, name="*")
app.update(sub_app_update)


if __name__ == "__main__":
    print(app.generate_docs())
```

->

---

# app

```console
app COMMAND
```

## Table of Contents

- [`update`](#app-update)

**Commands**:

* [`star`](#app-star)
* [`update`](#app-update)

## app update

```console
app update FOO
```

**Parameters**:

* `FOO, --foo`: **[required]**

---

As you can see there is no documentation for `app star` other than the entry under `Commands`.

Also, are the two forms (otherwise) equivalent or am I missing something?

BrianPugh:
## Render flattened subapp commands in generated docs

Commands flattened into a parent via `app.command(subapp, name="*")` showed up in the help-derived **Commands** panel but got no doc section or TOC entry (and a dead anchor link). Root cause: every docs format builds its command list through `iterate_commands()` in `cyclopts/docs/base.py`, which only walked `app._commands` and never `app._flattened_subapps`.

- `iterate_commands()` now merges flattened subapp commands using the same semantics as `_combined_meta_command_mapping()`/`resolved_commands()`: parent commands take precedence on name collision, and the subapp's own builtin `--help`/`--version` entries are skipped.
- Since markdown, RST, HTML, and the recursive doc walk all go through this one helper, the fix covers all formats, including nested flattening.

Fixes #904

### DIFF
--- a/cyclopts/docs/base.py
+++ b/cyclopts/docs/base.py
@@ -503,12 +503,16 @@ def iterate_commands(app: "App", include_hidden: bool = False, resolve_lazy: boo
     Tuple[str, App]
         (command_name, resolved_subapp) for each valid command.
     """
-    if not app._commands:
-        return
+    commands: dict[str, App | CommandSpec] = dict(app._commands)
+    # Merge commands from flattened subapps (registered via name="*"); parent commands take precedence.
+    for flattened in app._flattened_subapps:
+        for cmd_name in flattened:
+            if cmd_name not in commands and not _is_builtin_flag(flattened, cmd_name):
+                commands[cmd_name] = flattened._get_item(cmd_name, recurse_meta=False)
 
     seen: set[int] = set()
 
-    for name, app_or_spec in app._commands.items():
+    for name, app_or_spec in commands.items():
         if _is_builtin_flag(app, name):
             continue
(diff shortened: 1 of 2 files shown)


BrianPugh:
@stainbank can you give this a try? Thanks!

stainbank:
Looks good :)

PR #906, opened 2026-08-26
Handle a config file whose document root is empty or not a mapping

dylanpulver:
A `config.yaml` that is empty, or has everything commented out, kills the CLI:

```
$ myapp --x 5
Traceback (most recent call last):
  File "cyclopts/config/_common.py", line 56, in __call__
    config: dict[str, Any] = self.config.copy()
AttributeError: 'NoneType' object has no attribute 'copy'
```

`safe_load` returns `None` for those documents, as does `json.load` for a file holding `null`. `Toml` returns `{}` for its empty file and keeps working.

Behind it, `ConfigFromFile.config` is annotated `-> dict[str, Any]` but returns `None` on first access and `{}` on the second: the cache-hit branch at line 155 has `or {}`, the fresh-load branch at line 170 does not.

A scalar or list root gives the raw error that #873 removed for child nodes:

```
Yaml("list.yaml")   ->  AttributeError: 'list' object has no attribute 'items'
Yaml("scalar.yaml") ->  AttributeError: 'int' object has no attribute 'copy'
```

#873 put the mapping check inside the descent loop, and line 56 runs before the loop starts, so the root is the one node it never sees. Its three tests use `Dict({...})`, whose root is a dict by construction.

Both changes are in `_common.py`: line 170 becomes `return self._config or {}`, matching its sibling fifteen lines above, and the mapping check moves into `_ensure_mapping()` so it also runs on the root. The message for descended keys is unchanged, so #873's tests match it as written.

Deliberate: `or {}` maps a `[]` root to `{}` rather than to an error, which is what line 155 has done since 2025-01.

Tests cover empty and comments-only YAML, a non-mapping YAML root, a non-mapping `Dict` root with and without `root_keys`, and the first-versus-second read of `config`. Reverting only the guard fails 3 of the 6, reverting only `or {}` fails the other 3. Full suite 2613 passed, `pre-commit run --all-files` clean.

Written with Claude Code.

### DIFF
--- a/cyclopts/config/_common.py
+++ b/cyclopts/config/_common.py
@@ -47,26 +47,41 @@ class ConfigBase(ABC):
         """Return a string identifying the configuration source for error messages."""
         raise NotImplementedError
 
+    def _ensure_mapping(self, node: Any, traversed: list[str]) -> None:
+        """Raise a :class:`CycloptsError` if a visited configuration node is not a mapping.
+
+        Parameters
+        ----------
+        node: Any
+            The configuration node reached so far.
+        traversed: list[str]
+            Keys descended through to reach ``node``; empty for the document root.
+        """
+        if isinstance(node, dict):
+            return
+        keyword = "".join(f"[{k}]" for k in traversed)
+        location = f"key {keyword} " if keyword else ""
+        raise CycloptsError(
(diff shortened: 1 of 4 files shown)


BrianPugh:
thank you!

PR #909, opened 2026-08-27
register field-name CLI options when `validate_by_name` is set

### LINKED ISSUE #908: CLI rejects field-name options when a model uses `validate_by_name=True` (pydantic ≥ 2.11)
**Repo:** BrianPugh/cyclopts
**File:** `cyclopts/field_info.py` (line 174, in `_pydantic_field_infos`)
**Rule reference:** pydantic 2.11 split `validate_by_name` from `populate_by_name`

## Summary

When a pydantic model sets `validate_by_name=True` (pydantic ≥ 2.11, so the model **accepts its field names during validation**), cyclopts still decides which CLI option names to register by reading **only** `populate_by_name`:

```python
if pydantic_field.alias:
    if model.model_config.get("populate_by_name", False):
        names.append(python_name)
    names.append(pydantic_field.alias)
```

As a result, for aliased fields whose Python name is not a hyphen/underscore variant of the alias, the field-name CLI option is **rejected as "Unknown option"**, even though the underlying model would happily validate input by field name.

## Reproduction

cyclopts 4.23.3 + pydantic 2.11.10:

```python
from cyclopts import App
from pydantic import BaseModel, Field, ConfigDict

class Model(BaseModel):
    model_config = ConfigDict(validate_by_name=True)  # NOT populate_by_name
    full_name: str = Field(alias="userName")

app = App()

@app.default
def main(model: Model):
    ...

app(["--model.full_name", "Alice"])
```

Result:

```
╭─ Error ──────────────────────────────────────────────────────────────────────╮
│ Unknown option: --model.full_name. Did you mean --model.username?            │
╰──────────────────────────────────────────────────────────────────────────────╯
SystemExit: 1
```

while pydantic itself accepts the field name:

```python
Model.model_validate({"full_name": "Alice"})   # OK
```

and `--model.user-name` / `--model.username` parse fine. Only the **field-name** form — exactly what `validate_by_name=True` opts into — is missing.

## Expected vs actual

- **Expected:** a model configured with `validate_by_name=True` should accept its field names on the CLI (that is the whole point of the config), i.e. `--model.full_name` should parse.
- **Actual:** field-name options are only registered when `populate_by_name` is set; with the newer `validate_by_name` alone, they are rejected.

## Root cause

pydantic 2.11 split the old `populate_by_name` semantics: `populate_by_name` now covers name-based population (e.g. `construct`), while `validate_by_name` independently governs whether field names are accepted **during validation**. cyclopts' name-registration decision only reads the former, so it ignores the latter. This mirrors a bug polyfactory had with the same config key (fixed upstream by reading both).

## Suggested fix

Read both flags:

```python
if pydantic_field.alias:
    if model.model_config.get("validate_by_name", False) or model.model_config.get("populate_by_name", False):
        names.append(python_name)
    names.append(pydantic_field.alias)
```

Verified against cyclopts 4.23.3 + pydantic 2.11.10: with this one-line change, `--model.full_name` parses correctly, `--model.user-name` / `--model.username` still work, and models with only `populate_by_name=True` are unaffected. Full patch attached in `fix.patch`.

## Notes

- For simple aliases (`alias="firstName"` with field `first_name`), the failure is masked by cyclopts' fuzzy hyphen/underscore matching against the legacy lowercase alias form — the bug is visible whenever the field name is not a case/hyphen/underscore variant of the alias (e.g. `full_name` vs `userName`), or when the legacy alias form is undesirable.
- `validate_by_name` requires pydantic ≥ 2.11; cyclopts already requires `>=2.11.2`, so the key is always supported on the supported pydantic range.

xyf5432:
Fixes #908

## Summary

`_pydantic_field_infos` decides whether an aliased pydantic field's Python name is
registered as an acceptable CLI option by reading only
`model_config.get("populate_by_name")`. Since pydantic 2.11 split
`validate_by_name` out of that semantic, models that configure
`validate_by_name=True` (field names accepted during validation) get their
field-name CLI options rejected as "Unknown option" — while the model itself
accepts those names.

This mirrors the same bug in polyfactory, fixed upstream by reading both flags.

## Changes

- `cyclopts/field_info.py` (in `_pydantic_field_infos`):

  ```python
  - if model.model_config.get("populate_by_name", False):
  + if model.model_config.get("validate_by_name", False) or model.model_config.get("populate_by_name", False):
  ```

  No other name-registration path is touched.

## Testing

Verified with cyclopts 4.23.3 + pydantic 2.11.10:

- Model with `ConfigDict(validate_by_name=True)`, field `full_name: str = Field(alias="userName")`:
  - Before: `--model.full_name` → `Unknown option: --model.full_name. Did you mean --model.username?` (SystemExit 1).
  - After: `--model.full_name Alice` parses correctly.
  - Alias forms `--model.user-name` / `--model.username` continue to work in both cases.
- Model with `populate_by_name=True` (only): behavior unchanged.

### DIFF
--- a/cyclopts/field_info.py
+++ b/cyclopts/field_info.py
@@ -171,17 +171,31 @@ def _pydantic_field_infos(model) -> dict[str, FieldInfo]:
     for python_name, pydantic_field in model.model_fields.items():
         names = []
         if pydantic_field.alias:
-            if model.model_config.get("validate_by_name", False) or model.model_config.get("populate_by_name", False):
-                names.append(python_name)
-            names.append(pydantic_field.alias)
+            # Register a CLI name only if the model would actually accept it during
+            # validation (cyclopts builds the model via TypeAdapter.validate_python).
+            #   * validate_by_name  -> field names are accepted.
+            #   * validate_by_alias -> aliases are accepted (defaults to True).
+            # populate_by_name is the pre-2.11 spelling of validate_by_name; it is not
+            # recommended in pydantic 2.11+ and is slated for removal in pydantic v3, so
+            # we honor both across the supported pydantic range. pydantic forbids both
+            # validate_by_* being False, so at least one name is always registered.
+            validate_by_name = model.model_config.get("validate_by_name", False) or model.model_config.get(
+                "populate_by_name", False
+            )
+            validate_by_alias = model.model_config.get("validate_by_alias", True)
 
-            # Add legacy-compatible CLI form if not already present.
-            # This allows both "user-name" (new) and "username" (legacy) to work as CLI options.
-            # Old transform behavior: alias.lower() (no pascal_to_snake)
-            # New transform behavior: _pascal_to_snake(alias).lower()
-            legacy_form = pydantic_field.alias.lower()
-            if legacy_form not in names:
-                names.append(legacy_form)
+            if validate_by_name:
+                names.append(python_name)
+            if validate_by_alias:
+                names.append(pydantic_field.alias)
+
+                # Add legacy-compatible CLI form if not already present.
+                # This allows both "user-name" (new) and "username" (legacy) to work as CLI options.
+                # Old transform behavior: alias.lower() (no pascal_to_snake)
+                # New transform behavior: _pascal_to_snake(alias).lower()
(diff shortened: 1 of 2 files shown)


BrianPugh:
thank you!

PR #912, opened 2026-08-28
fix(docs): use module name for unnamed apps and drop duplicate root anchors

### LINKED ISSUE #910: Sphinx docs list the command as `sphinx-build` and generates duplicate labels
Cyclopts is making our CLI tools a lot more usable than our previous ones using plain `argparse`. Thanks for your work on this.

Just trying out the docs generation via Sphinx but finding that the command is mislabeled as `sphinx-build` and that all instances of the cyclopts directive get the same label, even if referring to different commands.

1. The command at the top of the docs is listed as `sphinx-build mycommand --output-path PATH ...` instead of `myapp mycommand --output-path PATH ...`. This can be fixed by adding `:usage-name: myapp`.
2. When using the `.. cyclopts::` directive in several places, we get warnings about `duplicate label cyclopts-sphinx-build` even if they relate to different commands and override the usage name. This makes it impossible to reference those sections of the docs.

`make-x.rst`:
```
.. cyclopts:: myapp.__main__:app
    :commands: make-x
    :usage-name: uv tool run myapp
```

`make-y.rst`:
```
.. cyclopts:: myapp.__main__:app
    :commands: make-y
    :usage-name: uv tool run myapp
```

Build command: `sphinx-build docs/ docs/_build/html` or `make -C docs html` or `uv run --group docs make -C docs html`.

Warning:
```
/home/me/myapp/docs/make-y.rst:2: WARNING: duplicate label cyclopts-sphinx-build, other instance in /home/me/myapp/docs/make-x.rst
```

BrianPugh:
Fixes #910 

@carlfischerjba can you please give this a try? Thanks!

### DIFF
--- a/cyclopts/docs/base.py
+++ b/cyclopts/docs/base.py
@@ -327,49 +327,20 @@ def build_command_chain(command_chain: list[str] | None, command_name: str, app_
         return [app_name, command_name]
 
 
-def apply_usage_name(command_chain: list[str], usage_name: str | None) -> list[str]:
-    """Return a display command chain with the root replaced by ``usage_name``.
-
-    When ``usage_name`` is ``None``, returns ``command_chain`` unchanged so callers
-    can use this helper unconditionally. When ``usage_name`` is ``""``, the root
-    token is dropped rather than substituted, so downstream formatters never see
-    an empty element (which would render as stray leading/internal whitespace).
-    When the chain is empty and ``usage_name`` is a non-empty string, returns a
-    single-element list containing ``usage_name``.
-
-    Parameters
-    ----------
-    command_chain : list[str]
-        The logical command chain (root app name first).
-    usage_name : str | None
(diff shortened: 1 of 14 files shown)


carlfischerjba:
Thank you @BrianPugh for a very prompt and thorough fix. It works. No warnings during the docs build, the app name is correct in the docs, and the span id/label contains the correct app name and command name.

I hadn't realised that I could have partially worked around this by simply giving the app a name, although the duplicate labels remained because previously there was one for the app in addition to one for the app + command. Your fix seems to have removed the former which wouldn't have been needed.

If I add several `cyclopts` directives for the same app and command, I get a `WARNING: Duplicate explicit target name: "cyclopts-myapp-mycommand"` but that's not going to be a problem for me in practice.

BrianPugh:
I think I can address that in a follow up PR, thanks!

PR #916, opened 2026-08-29
feat(help): make help colors themeable via cyclopts.* named styles

BrianPugh:
Makes every color in the default (Rich) help output overridable through a named `cyclopts.*` [Rich style](https://rich.readthedocs.io/en/stable/style.html#style-themes), and fixes the low-contrast `dim` defaults that #914 reported.

Users restyle any element by defining that key in their Console's `rich.theme.Theme`:

```python
from rich.console import Console
from rich.theme import Theme
from cyclopts import App

app = App(console=Console(theme=Theme({"cyclopts.required": "bold yellow"})))
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #920, opened 2026-09-01
Edit documentation of color styles

stanwest:
This PR builds on #916 with several changes intended to improve readability of the new documentation of the named styles. The changes are as follows:

*   In the table of styles:
    *   Move the column of default values to the end.
    *   In the columns of names and defaults, add quotation marks to string literals. (This matches how string literals appear in the [table](https://github.com/BrianPugh/cyclopts/blob/c8df0e40e786732c2915f40fbd9628a4047cc49c/docs/source/api.rst?plain=1#L654) of values of `App.result_action`.
    *   In the "Applies to" column, capitalize and punctuate as phrases rather than sentences, change literal text to quoted text, and clarify the asterisk marker.
*   Edit a subsequent reference to the prefix.

The table appears as follows before and after this PR, respectively:

Color style table, previous

Color style table, revised

A further improvement would be to reorder the rows of the table either alphabetically by style name or by order of appearance in rendered help (i.e., usage, border, name, etc.).

### DIFF
--- a/docs/source/help_customization.rst
+++ b/docs/source/help_customization.rst
@@ -291,35 +291,35 @@ element; keys you don't define fall back to the built-in default.
 
 .. list-table::
    :header-rows: 1
-   :widths: 30 20 50
+   :widths: 30 50 20
 
    * - Style name
-     - Default
      - Applies to
-   * - ``cyclopts.name``
-     - ``cyan``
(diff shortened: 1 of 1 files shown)


BrianPugh:
thanks!

PR #935, opened 2026-09-07
fix: honor subcommand-level configuration (end_of_options_delimiter, error settings, backend, result_action)

### LINKED ISSUE #933: Bug report: Sub-command `end_of_options_delimiter` ignored since #524 (4.0.0)
Example program:

```python
from typing import Annotated

app = App(version_flags=[])


@app.command(help_flags=[], version_flags=[], end_of_options_delimiter="")
def k(*args: Annotated[str, Parameter(allow_leading_hyphen=True)]):
    """Pass-through (e.g. to kubectl): the literal `--` must survive."""
    print(list(args))

app(["k", "exec", "pod", "--", "sh", "-c", "id"])
```

Before #524:

```
['exec', 'pod', '--', 'sh', '-c', 'id']
```

After #524:

```
# Notice `--` is still stripped even with end_of_options_delimiter=""
['exec', 'pod', 'sh', '-c', 'id']
```

BrianPugh:
Fixes #933. Stacked on #934 (base will retarget to `main` once #934 merges).

## The bug (one root cause, four symptoms)

`AppStack` is **per-app**: entering the command context pushes the full chain `[root, sub]` onto the *subcommand's* stack but only `[root]` onto the *root's* stack. Several top-level orchestration methods on the *entry* app resolved command-scoped configuration via `self.app_stack` (the root/entry app), which can never see the invoked subcommand — so a subcommand-level setting was silently ignored and the root's value (or a default) was used instead.

This surfaced as four distinct regressions, all the same root cause:

| Setting | Symptom when set on a subcommand |
|---|---|
| `end_of_options_delimiter` | ignored; `--` was still stripped (the reported #933 — breaks pass-through commands) |
| `print_error` / `exit_on_error` / `help_on_error` / `verbose` / `error_formatter` | ignored; a subcommand's `exit_on_error=False` still exited, its `error_formatter` never ran |
| `backend` | ignored; a subcommand's `backend="trio"` ran under asyncio |
| `result_action` | ignored; a subcommand's `result_action` was never applied |

None of these leak upward or sideways — a subcommand's setting governs only that subcommand's invocation, with parents supplying defaults (standard child-overrides-parent).

## The fix

The rule: **resolve command-scoped config from the invoked command's `app_stack`**, which equals `self.app_stack` only when `self` is that command.

- `end_of_options_delimiter` — resolve from `command_app.app_stack` in `_parse_known_args` (alongside `_config`, which already did).
- error-reporting settings — resolved inside `_parse_known_args` while the invoked command's context (full chain + propagated overrides) is live, then carried to `_handle_parse_error` on the exception (a private `CycloptsError._error_report_settings`), since the handler runs after the context is torn down. Falls back to `self.app_stack` for errors raised before a command is resolved (e.g. tokenization), where the root scope is correct.
- `backend` / `result_action` — factor the "which command am I running" logic (command-chain resolution + help/version trimming) out of `_parse_known_args` into `_resolve_invoked_command`, and use it at the execution sites (`__call__`, `run_async`, `interactive_shell`) to resolve from the invoked command. `_handle_result_action` now scopes both `result_action` and the output console to the invoked command.

## Why `self.app_stack` is still used elsewhere

`self.app_stack` remains correct wherever `self` already *is* the app whose config is wanted — the property accessors (`config`, `console`, `error_console`, `version_format`, `default_parameter`) and the `version_print`/`help_print` commands bound to the invoked command — and as the fallback when no subcommand is involved (single-app use, or pre-command errors). The per-app design is fine; only the entry-app orchestration sites were resolving from the wrong app.

## Tests

Each fix has a failing-test-first / fix / coverage structure. New guards cover: subcommand `end_of_options_delimiter` (the repro, plus a custom delimiter, child-overrides-parent, and a nested grandchild); subcommand `error_formatter` / `exit_on_error` / `print_error` via both `__call__` and `parse_args`; subcommand `backend="trio"` and `result_action` via both `__call__` and `run_async`, plus a sibling-isolation guard. Existing suites only ever set these on the root app. Full suite green (2761 passed).

### DIFF
--- a/cyclopts/core.py
+++ b/cyclopts/core.py
@@ -132,6 +132,16 @@ def _get_version_command(app):
         return app.version_print
 
 
+#: Attributes controlling how a parse error is reported. Resolved together so the invoked
+#: command's scope is applied consistently (see ``App._handle_parse_error``).
+_ERROR_REPORT_ATTRIBUTES = ("print_error", "exit_on_error", "help_on_error", "verbose", "error_formatter")
+
+
+def _resolve_error_report_settings(app_stack: AppStack) -> dict[str, Any]:
+    """Resolve the error-reporting settings from ``app_stack`` (unset values stay ``None``)."""
+    return {name: app_stack.resolve(name) for name in _ERROR_REPORT_ATTRIBUTES}
+
+
 def _apply_parent_defaults_to_app(app: "App", parent_app: "App") -> None:
     """Apply parent app's group defaults to app if not already set.
 
@@ -1744,6 +1754,12 @@ class App:
                     e.command_chain = command_chain
                 if e.console is None:
                     e.console = command_app.error_console
+                # Resolve error-reporting settings now, while the invoked command's context
+                # (full command chain + propagated overrides) is live on ``command_app.app_stack``.
+                # ``_handle_parse_error`` runs after this context is torn down, where ``self`` (the
+                # entry app) can no longer see subcommand-level settings (issue #933 family).
+                if e._error_report_settings is None:
+                    e._error_report_settings = _resolve_error_report_settings(command_app.app_stack)
                 raise
 
         return command, bound, unused_tokens, ignored, argument_collection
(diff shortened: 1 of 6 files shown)


BrianPugh:
Superseded by #936. This PR was auto-closed by GitHub when #934 (the fix it was stacked on) merged and its base branch was deleted, and it can't be reopened because the head branch was rebased afterward. #936 is the same work, rebased onto `main`.

PR #939, opened 2026-09-08
v5: rebuild interactive_shell on cmd.Cmd

BrianPugh:
## Rebuild `App.interactive_shell` on `cmd.Cmd`

Targets the `v5-develop-interactive-shell` staging branch, which collects all shell PRs before one merge into `v5-develop`. First of three (next: `intro`/`history_file`/`"exit"` quit word; then tab completion once `feat-dynamic-completer` lands). This one is a structural move with two small breaking changes, no new features.

- New private module `cyclopts/_shell.py` holds a `cmd.Cmd` subclass. `onecmd` is overridden wholesale, so `cmd`'s own `do_*`/`help`/`?`/`!` dispatch can never steal a user command. `cmd.Cmd` buys us the readline completer install/restore hook that the completion PR needs.
- `completekey=""` so tab keeps readline's default behavior until the completion PR wires it up.
- Ctrl-C handling stays outside `cmdloop` because `cmd.Cmd` does not catch it. A `KeyboardInterrupt` re-raised from a command (when `suppress_keyboard_interrupt=False`) is flagged so the outer loop does not mistake it for Ctrl-C at the prompt.
- **Breaking:** parameters after `prompt` are now keyword-only. Only public `App` method that lacked the `*`.
- **Breaking:** `import readline` moved from `cyclopts/core.py` module level into `_shell.py`, which is imported lazily. Programs that only import cyclopts no longer get `input()` altered.
- Tests now patch `builtins.input` and `cyclopts._shell.readline` instead of `cyclopts.core.*`.

### DIFF
--- a/cyclopts/_shell.py
+++ b/cyclopts/_shell.py
@@ -1,0 +1,114 @@
+"""Interactive shell backing :meth:`cyclopts.App.interactive_shell`.
+
+Imported lazily: ``readline`` mutates :func:`input` for the whole process, so programs
+that merely import cyclopts should not pay for it.
+"""
+
+import cmd
+import traceback
+from collections.abc import Collection
+from typing import TYPE_CHECKING, Any
+
+from cyclopts.exceptions import CycloptsError
+
+try:
+    # Makes arrow keys and history work inside the shell.
+    import readline
+except ImportError:  # pragma: no cover
+    # Not available on windows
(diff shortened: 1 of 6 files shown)


BrianPugh:
Superseded by the minimal version; cmd.Cmd and the module split didn't pull their weight.

PR #940, opened 2026-09-08
v5: move interactive_shell loop into _shell.py (alternative to #939)

BrianPugh:
## Move `interactive_shell` loop into `cyclopts/_shell.py`

Alternative to #939 for independent comparison. Same two breaking changes, same test updates, but the original hand-written loop is kept as a plain function instead of subclassing `cmd.Cmd`. Targets the `v5-develop-interactive-shell` staging branch.

- `run_shell()` in the new lazily imported `cyclopts/_shell.py` is the old loop body verbatim. `App.interactive_shell` keeps only argument normalization and the `app_stack` override context.
- **Breaking:** parameters after `prompt` are now keyword-only, the only public `App` method that lacked the `*`.
- **Breaking:** `import readline` moved out of `cyclopts/core.py` module level into `_shell.py`. Programs that only import cyclopts no longer get `input()` altered or pay the import cost.
- Tests now patch `builtins.input` and `cyclopts._shell.readline` instead of `cyclopts.core.*`.

Versus #939: no `EOF` sentinel word, no `interrupted_in_command` flag, no `run()`/`cmdloop()` retry wrapper. The completion PR will call `readline.set_completer` directly, which is under ten lines either way.

### DIFF
--- a/cyclopts/_shell.py
+++ b/cyclopts/_shell.py
@@ -1,0 +1,87 @@
+"""Interactive shell backing :meth:`cyclopts.App.interactive_shell`.
+
+Imported lazily: ``readline`` mutates :func:`input` for the whole process, so programs
+that merely import cyclopts should not pay for it.
+"""
+
+import traceback
+from collections.abc import Collection
+from typing import TYPE_CHECKING, Any
+
+from cyclopts.exceptions import CycloptsError
+
+try:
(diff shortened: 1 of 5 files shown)


BrianPugh:
Superseded by the minimal version.

PR #942, opened 2026-09-08
fix: reject NaN in numeric range validators

kudala-bharani:
`NaN` currently passes every `Number` range check. For example, this returns without an error:

```python
from cyclopts.validators import Number

Number(gte=0, lte=1)(float, float("nan"))
```

The same happens when parsing `nan` for a `NormFloat` CLI parameter. The checks reject values using the opposite comparison, but comparisons with NaN are false in either direction.

This negates each required comparison instead. It keeps the existing error messages and still permits infinities where the bound allows them.

Added tests for all four bounds, scalar and nested-container inputs, and the six constrained float aliases. All 18 NaN regression cases fail before the fix; four additional cases cover allowed infinities.

Validation on Python 3.13.14:

- Focused numeric tests: 44 passed.
- Full suite with coverage: 2,740 passed, 56 skipped, 12 deselected; 24 snapshots passed.
- All pre-commit checks passed, including Ruff and Pyright.

### DIFF
--- a/cyclopts/validators/_number.py
+++ b/cyclopts/validators/_number.py
@@ -69,16 +69,17 @@ class Number:
             if not isinstance(value, int | float):
                 return
 
-            if self.lt is not None and value >= self.lt:
+            # Negate the required comparison so NaN cannot bypass a bound.
+            if self.lt is not None and not value < self.lt:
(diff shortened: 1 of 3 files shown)


BrianPugh:
thank you!

Issue #885, opened 2026-08-08
Add support for metavars


Rikiub:
Currently there is not API for handle and change the generated metavars, neither in [HelpEntry](https://cyclopts.readthedocs.io/en/stable/api.html#cyclopts.help.HelpEntry) or [Parameter](https://cyclopts.readthedocs.io/en/stable/api.html#cyclopts.Parameter) APIs.

You can take as example how [Typer implement this feature](https://typer.tiangolo.com/tutorial/arguments/help/?h=metavar#custom-help-name-metavar).

Would be great have this kind of feature, it could be more easy to customize the help panels.

BrianPugh:
I'm open to it! Just to guide me, can you give a real life example of where you would want to use this feature? Naively, it seems kind of neat, but not actually useful?

Rikiub:
## Explanation

Maybe a custom metavar is not useful with the default structure:

```
╭─ Filters ─────────────────────────────────────────────────────────────────────────────────────────╮
│ TYPE --type -t                      Type of stream to download (downloads best format by          │
│                                     default). [choices: video, audio, video-only, audio-only]     │
│ QUALITY --quality -q                Prefered target quality. [choices: 144, 240, 360, 480, 720,   │
│                                     1080]                                                         │
│ VIDEO-CODEC --video-codec --vcodec  Prefered video codec.                                         │
│ AUDIO-CODEC --audio-codec --acodec  Prefered audio codec.                                         │
╰───────────────────────────────────────────────────────────────────────────────────────────────────╯
```

But in case I want implement my own `HelpFormatter`, it would be something like this:

```
╭─ Filters ─────────────────────────────────────────────────────────────────────────────────────────╮
│ --type -t TYPE                      Type of stream to download (downloads best format by        │
│                                       default). [choices: video, audio, video-only, audio-only]   │
│ --quality -q QUALITY                Prefered target quality. [choices: 144, 240, 360, 480, 720, │
│                                       1080]                                                       │
│ --video-codec --vcodec VIDEO-CODEC  Prefered video codec.                                       │
│ --audio-codec --acodec AUDIO-CODEC  Prefered audio codec.                                       │
╰───────────────────────────────────────────────────────────────────────────────────────────────────╯
```

Looks fine, but It can be better if I could replace the long `VIDEO-CODEC` metavar with just `CODEC`. More examples:

- `--audio-codec AUDIO-CODEC` -> `--audio-codec CODEC`
[...]

PR #949, opened 2026-09-12
fix(help): show CHOICE metavar for choice parameters and collapse variadic-collection metavars to X...

BrianPugh:
Addresses both items from Rikiub's feedback on the metavar feature in [#885](https://github.com/BrianPugh/cyclopts/issues/885#issuecomment-5643310179).

## 1. Show `CHOICE` metavar for enum / `Literal` / explicit-choices parameters

Enum, `Literal`, and explicit-`Parameter.choices` parameters previously rendered a *blank* value placeholder in help panels (relying on the adjacent `[choices: ...]` list) while every other value-consuming type showed its type name (`PATH`, `STR`, `INT`, ...). This made choice parameters look inconsistent with the rest of the panel — and with the usage line, which already showed `CHOICE`.

Panels now show `CHOICE`, consistent with the usage line and every other type:

```
│ *  --codec-req CHOICE     required keyword enum [choices: av1, vp9, h264]    │
│    --codec CHOICE         enum via | None [choices: av1, vp9, h264]          │
│    --mode CHOICE          Literal via | None [choices: fast, slow]           │
│    --pick CHOICE          explicit Parameter.choices [choices: a, b]         │
```

Since panel and usage lines now render metavars identically, the vestigial `in_usage` suppression path and the `choice` threading through `_type_metavar` are removed.

## 2. Collapse variadic-collection metavars to `X...`

`list`, `set`, `frozenset`, `Sequence`, `Iterable`, and the other single-element iterables consume one value per CLI token — exactly like `tuple[X, ...]` — but rendered as `LIST[STR]` / `SET[STR]` / `SEQUENCE[STR]` / `ITERABLE[STR]` while `tuple[str, ...]` rendered `STR...`. Rikiub flagged this inconsistency directly.

A single-token element now collapses to `X...` for every variadic collection, matching `tuple`:

| type | before | after |
| --- | --- | --- |
| `list[str]` | `LIST[STR]` | `STR...` |
| `set[int]` | `SET[INT]` | `INT...` |
| `Sequence[str]` | `SEQUENCE[STR]` | `STR...` |
| `list[Literal["a","b"]]` | `LIST[CHOICE]` | `CHOICE...` |

A **multi-token** element keeps the `LIST[...]` form so its grouping stays unambiguous — `list[tuple[int, str]]` → `LIST[INT STR]`, not the ambiguous `INT STR...`; `list[tuple[int, ...]]` → `LIST[INT...]`. `dict` is unaffected (`DICT[STR, INT]`).

## 3. Recognize abstract collections consistently

The subsystems that reason about "is this a collection" disagreed on the abstract-collection ABCs. `_convert.py` normalizes `collections.abc.Collection` / `MutableSequence` / `Set` / ... to their concrete types (via `_abstract_to_concrete_type_mapping`), but the choices extractor and the shell-completion action detector both gated on plain `ITERABLE_TYPES` membership, so those abstract hints were silently missed:

- `Collection[Literal["a", "b"]]` advertised the `CHOICE` metavar but dropped its `[choices: a, b]` list.
- `Collection[Path]` (and the other abstract collections) offered no file completion.

Both now recognize the same set of collections the converter does, so an abstract hint behaves exactly like its concrete counterpart everywhere.

## 4. Drop speculative `Container` support (breaking)

`collections.abc.Container` was in the abstract-to-concrete mapping as a completeness follow-up (never user-requested). Unlike every other mapped ABC, `Container` only promises `__contains__` — it is *not* an `Iterable` subclass — so it fit neither the "collection of values" model the parser assumes nor the `Iterable`-based checks the rest of the code relies on, and it was the sole reason those checks needed special-casing.

`Container[...]` now raises (surfaced as a `CoercionError`) instead of quietly coercing to a `list`. With it gone, every remaining collection origin is an `Iterable` subclass, so the choices, metavar, and completion sites all reduce to a single `is_iterable_type` check — no per-type special cases. This only ever shipped in the v5 pre-releases, so no released version is affected.

## Notes

- Positional-capable rows are unaffected — they still show their name-derived identifier (`CODEC --codec`), not the metavar.

## Tests & docs

- Updated assertion tests that encoded the old behavior (`test_help_metavar.py`, `test_help.py`, `test_bind_generic_class.py`, `test_burgery.py`).
- Added regression coverage for abstract-collection choices/completion (`test_annotations.py`, `test_behavior.py`) and for the deliberate `Container` exclusion (`test_coercion.py`).
- Regenerated syrupy docs snapshots — all diffs are the metavar changes above.
- Docs updated (`help_customization.rst`, `api.rst`): dropped the obsolete panel-suppression explanation and documented the variadic-collection form.

Full suite: 3431 passed, 6 skipped.

https://claude.ai/code/session_01QecYSBdMtW2ATDcV7Ac4ep

### DIFF
--- a/cyclopts/_convert.py
+++ b/cyclopts/_convert.py
@@ -70,7 +70,6 @@ _abstract_to_concrete_type_mapping: dict[type, type] = {
     collections.abc.MutableSet: set,
     collections.abc.MutableSequence: list,
     collections.abc.Collection: list,
-    collections.abc.Container: list,
     collections.abc.Reversible: list,
     collections.abc.Mapping: dict,
     collections.abc.MutableMapping: dict,
--- a/cyclopts/annotations.py
+++ b/cyclopts/annotations.py
@@ -355,7 +355,7 @@ def get_choices_from_hint(type_: Any, name_transform: Callable[[str], str]) -> l
                 choices.extend(x)
     elif _origin is Literal:
         choices.extend(str(x) for x in get_args(type_))
-    elif _origin in ITERABLE_TYPES:
+    elif _origin in ITERABLE_TYPES or _origin in VARIADIC_COLLECTION_TYPES:
         args = get_args(type_)
         if len(args) == 1 or (_origin is tuple and len(args) == 2 and args[1] is Ellipsis):
             choices.extend(get_choices(args[0]))
--- a/cyclopts/annotations.py
+++ b/cyclopts/annotations.py
@@ -1,7 +1,17 @@
 import inspect
 import sys
 import typing
-from collections.abc import Callable, Iterable, Sequence
+from collections.abc import (
+    Callable,
+    Collection,
+    Container,
+    Iterable,
+    MutableSequence,
+    MutableSet,
+    Reversible,
+    Sequence,
+    Set,
+)
 from enum import Enum, Flag
 from functools import partial
 from types import UnionType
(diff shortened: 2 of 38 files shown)


BrianPugh:
@Rubikoid can you give this a try? I think it addresses both your feedback items!

Rubikoid:
Oh wow hello! I guess you intended to mention @Rikiub :)

BrianPugh:
Whoops! Sorry about that Rubikoid! Thanks! 😅

Rikiub:
Hello again! Looks like the problem is fixed, now is working like should 👍

PR #951, opened 2026-09-14
Validate Fraction and Decimal values

### LINKED ISSUE #950: Number validator always passes numeric types other than int and float
The `cyclopts.validators.Number` validator passes essentially all types other than `int`, `float`, and some iterables (including `str`). This prevents it from working usefully for other representations of real numbers, such as `decimal.Decimal`, `fractions.Fraction`, and *numpy* scalars.

For example, *cyclopts* does not validate the parameter in this script:

>   ```python
>   from decimal import Decimal
>   from typing import Annotated
>   
>   from cyclopts import App
>   from cyclopts.parameter import Parameter
>   from cyclopts.validators import Number
>   
>   app = App()
>   
>   
>   @app.default()
>   def main(cost: Annotated[Decimal, Parameter(validator=Number(gt=0))]) -> None:
>       """Record a purchase."""
>       print(f"Cost: {cost}.")
>   
>   
>   if __name__ == "__main__":
>       app()
>   ```

The desired behavior is to reject a non-positive cost, but, given the argument "-100.00", instead the program proceeds and prints:

>   Cost: -100.00.

stanwest:
Fixes #950.

The first two commits test and resolve the issue. The first commit introduces tests that validate Fraction and Decimal values; the validations that should raise `ValueError` fail, and their tests are marked as expected failures. The second commit resolves the issue and removes the markers for expected failures.

The remaining commits attempt to improve the organization of the test module. One could further improve the module by removing redundant tests and testing the four comparisons more equally, but the final state here just incorporates the prior tests as parameterized test inputs.


BrianPugh (approved):

BrianPugh:
@stanwest can you cherry-pick 70c45f61f3fb88dc00d4efde68b647a8037b0504 from my `fix-decimal-nan-number-validator` branch? I think (if you think that commit is good) after that we can merge in this PR and do a patch release.

stanwest:
> @stanwest can you cherry-pick [70c45f6](https://github.com/BrianPugh/cyclopts/commit/70c45f61f3fb88dc00d4efde68b647a8037b0504) from my `fix-decimal-nan-number-validator` branch?

@BrianPugh, thanks for evaluating this PR. I've thought about adding a check for `NaN` values as in https://github.com/BrianPugh/cyclopts/commit/70c45f61f3fb88dc00d4efde68b647a8037b0504. To me, it seems best for `Number` to avoid explicitly checking for `NaN` values (to keep it relatively simple and type-agnostic) and for tests of `Number` involving `Decimal` `NaN` values to emulate application code and set an appropriate decimal context.

To elaborate, if `Number` were to check for `NaN` values, I'd be concerned that the check `is_nan = value != value` in 70c45f61f3fb88dc00d4efde68b647a8037b0504 is somewhat brittle: Ruff's *pylint* rule "comparison-with-itself" (PLR0124) warns about it if enabled, and whether it detects a `NaN` in other types (i.e., a custom class or extension type) depends on that type's `__ne__` behavior. However, because the preferred way to detect a `NaN` varies by type—`math.isnan(_)` for `float`, `_.is_nan()` for `Decimal`, and `numpy.isnan(_)` for *numpy* types—`Number` would need to branch on the type of value, and it's unclear how to handle an arbitrary type. If `Number` were to go that far, should it also check for finiteness, which likewise varies by type? (Presently, one can supply "inf" on the command line where a `float` parameter is required, and `Number(gte=0)(float, float("inf"))` will let it pass.)

Alternatively, if `Number` limits its operations to comparisons and the modulo operator, it remains more flexible about types. Because the `validator` parameter of the `Parameter` class allows a sequence of validators, developers can place checks for `NaN` values or finiteness in that sequence ahead of `Number`, and those checks can be appropriate for the type declared for that parameter. That composable approach seems better to me than trying to make one numerical validator handle many checks and types.

As for the test suite, would you consider it better to apply the decimal context only to tests that specifically need it? In this PR, that would be a single test, namely the one involving `Decimal("NaN")`. The decimal context could be applied immediately around the `validator(type_, value)` call as it is now, or it could be moved outward into a test fixture, which might better embody how application code might set the context more broadly. I'm willing to offer either of those changes.

BrianPugh:
Thanks for the detailed response! The nan-handling was a [recent addition](https://github.com/BrianPugh/cyclopts/pull/942), so its certainly something I want to support. Sorry for some terse answers here, because I have to run somewhere!

> developers can place checks for NaN values or finiteness in that sequence ahead of Number

True, but I think that NaN-checking is important and technically correct within the `Number` validator. This doesn't exactly need to be infinitely extendable, the number of Number-like objects that one could expect to input from a CLI is a small, finite set.

I agree with the points you made; and I'm not married to any specific implementation, but I do think we need proper NaN-handling in this validator.

> As for the test suite, would you consider it better to apply the decimal context only to tests that specifically need it?

Whatever is cleaner/terser, I have no strong preference.

stanwest:
> I agree with the points you made; and I'm not married to any specific implementation, but I do think we need proper NaN-handling in this validator.

I agree that that would be best.

At the moment, I see a couple of tricky aspects to equipping `Number` to validate `Decimal` values.  The first is that, if one wants `Number` to avoid raising `InvalidOperation`, the expression `is_nan = value != value` in 70c45f61f3fb88dc00d4efde68b647a8037b0504 is not reliably quiet in that sense; under `decimal`'s `BasicContext` and `DefaultContext`, for example, it will raise `InvalidOperation` when `value` is a signaling NaN, `Decimal("snan")`. Even in a context with the corresponding trap disabled, the expression will set the `InvalidOperation` flag in the context as a side effect. However, the `Decimal.is_nan()` method appears to be a reliably quiet way to detect a NaN, as in:

>   ```python
>   is_nan = (
>       math.isnan(value)
>       if isinstance(value, float)
>       else value.is_nan()
>       if isinstance(value, Decimal)
>       else (value != value)
>   )
>   ```

The second tricky aspect is that the `Number` validator is presently annotated to accept a `float` (or `int`) as the modulo divisor, but the modulo operator raises `TypeError` when given one `Decimal` and one `float` argument (in either order); it does not support the operation between those types. One way to avoid the issue is to have a distinct validator for `Decimal` and possibly each of the other built-in numeric types. (*Pydantic* [takes this approach](https://pydantic.dev/docs/validation/latest/api/pydantic/standard_library_types/#numbers).) Each validator could have keyword arguments appropriate for its type; e.g., `float` and `Decimal` can check for finiteness, whereas `int` and `Fraction` need no such check. Those arguments could be typed appropriately for the expected value type; in particular, one might prefer to avoid checking `Fraction` or `Decimal` instances against `float` thresholds or modulo divisors.

BrianPugh:
My general philosophy here is that I'm happy to absorb extra complexity *inside* Cyclopts if it results in a simpler, more intuitive experience for the user. For that reason, I would like there to only be a **single `Number` validator** rather than splitting into per-type validators the way pydantic does. The set of number-like types someone can realistically type on a command line is small and finite (`int`, `float`, `Decimal`, `Fraction`), so I'd rather one validator handle all of them than make users pick `NumberFloat` vs. `NumberDecimal` etc. and know which knobs each supports.

Onto the points you brought up:

## NaN detection
Agreed, let's use your proposed nan-detector; it's strictly better. I'd be tempted to go as far as by checking for existence of methods `isnan`/`is_nan` instead of type, but I think it's probably splitting hairs at this point.

## Modulo cross-type
I don't think this forces us into separate validators. The comparisons already work across `Decimal`/`Fraction` and `float` thresholds; the *only* operation that blows up is `Decimal % float` (`Fraction % float` is fine and returns a `float`). So we can special-case just that one spot — coerce the divisor to the value's type when the value is a `Decimal`:

```python
divisor = self.modulo
if isinstance(value, Decimal) and not isinstance(divisor, Decimal):
    # str() to avoid this nonsense:
    # >>> Decimal(0.1)
    # Decimal('0.1000000000000000055511151231257827021181583404541015625')
    # >>> Decimal(str(0.1))
    # Decimal('0.1')
    divisor = Decimal(str(divisor))

if value % divisor:
    raise ValueError(...)
```

That keeps the `int | float` bound/divisor annotations as-is and contains the wart to a couple of lines, which I much prefer over a validator-per-type API.

stanwest:
> My general philosophy here is that I'm happy to absorb extra complexity _inside_ Cyclopts if it results in a simpler, more intuitive experience for the user. For that reason, I would like there to only be a **single `Number` validator** rather than splitting into per-type validators the way pydantic does.

I expect that that desire will be in tension against the nature of Python's built-in numeric types and possibly other numeric types. A developer using *cyclopts* to handle numbers as `Decimal` may well want to compare against `Decimal` thresholds and use a `Decimal` modulo divisor to stay within the guarantees of the decimal system, and likewise with the `Fraction` system. However, doing so is excluded by `Number`'s current type annotations (which I understand you want to maintain), and, as we've seen, `Decimal` values arithmetically interoperate with neither `float` nor `Fraction`.

> So we can special-case just that one spot — coerce the divisor to the value's type when the value is a `Decimal`:

I advise against that approach. Coercing the divisor through `str` generally can change its value, and I wouldn't want that in an application where numerical precision, accuracy, and/or consistency are crucial.

If allowing `Number`'s thresholds and divisor to be subclasses of either `int` or `float` is important, it may be best not to claim support for `Decimal` and `Fraction` values but rather steer developers who want to use those types to write appropriate validators.

BrianPugh:
> However, doing so is excluded by Number's current type annotations (which I understand you want to maintain)

I think it would be totally fine to expand the type annotations of the `Number` validator attributes; e.g.
```python
lt: Decimal | Fraction | int | float | None = None
```

that way, anyone in those number class systems can specify all checks without unexpected results. If this is the limiting factor, by all means lets expand the type annotation for all the attributes!

PR #952, opened 2026-09-14
fix: validate Decimal, Fraction, and other numbers.Number types in the Number validator

### LINKED ISSUE #950: Number validator always passes numeric types other than int and float

BrianPugh:
Closes #950.

`cyclopts.validators.Number` silently accepted every value that wasn't an `int` or `float`. The scalar branch gated on `isinstance(value, int | float)` and `return`ed for anything else, so `decimal.Decimal`, `fractions.Fraction`, and numpy scalars skipped range validation entirely — `Number(gt=0)` happily let a `Decimal("-100")` through.

## Fix

Broaden the gate to `numbers.Number`. Note that `Decimal` is deliberately *not* registered as `numbers.Real` in CPython, so `numbers.Real` would have missed the issue's headline `Decimal` case; `numbers.Number` covers `Decimal`, `Fraction`, and numpy scalars. `complex` now passes the gate too and raises `TypeError` on `<`, which is consistent with the existing str behavior (range-checking a complex number is nonsensical).

The NaN guard also changes. The old negation trick relied on `float('nan') < x` returning `False`, but `Decimal('nan') < x` *raises* `InvalidOperation`. It now uses a type-agnostic `value != value` check (true only for NaN across float/Decimal/numpy, and never raises) that short-circuits before the ordered comparison, so Decimal/numpy NaN get the same clean rejection as float NaN.

## Tests

Added parametrized Decimal/Fraction validation and a Decimal-NaN rejection test. Full validators + number-types suites pass (100 passed).

https://claude.ai/code/session_01ThV97mheYfZvAryRXjhr9y


stanwest:
> Broaden the gate to `numbers.Number`.

See also #951 for an alternative, simpler solution. I'd rather see this validator limit its concern to whether its comparisons produce a truthy result than with the types of values.

> The NaN guard also changes. The old negation trick relied on `float('nan') < x` returning `False`, but `Decimal('nan') < x` _raises_ `InvalidOperation`.

In particular, note that the raising of `InvalidOperation` by operations such as `Decimal('nan') < x` is by design with the default decimal context. It may be better to leave it to the developer using *cyclopts* to explicitly set the decimal context that they desire for their application.

BrianPugh:
> I'd rather see this validator limit its concern to whether its comparisons produce a truthy result than with the types of values.

I think thats a good general guiding principal. I'll review/compare this to #951 this evening. I haven't even reviewed this PR (my agent just made it so i can just review when i get a chance). I'm leaning towards your solution, but i just need a chance to dive deeper into it. Thanks!

BrianPugh:
closing in favor of #951

